PR #16, opened 2023-06-23
Fix: DCL mode creates OpenAiStreamClient singleton and omits volatile keyword.

novohit:


openai0229:
Thank you for the concurrency fix. The legacy OpenAIClient singleton and CeOpenAiStreamClient path changed by this pull request no longer exist in the current Community AI architecture. The patch is therefore no longer applicable and will be closed.

PR #42, opened 2023-06-25
Update README_CN.md

zsinba:
单词拼写错误


openai0229:
Thank you for catching the documentation issue. The current Chinese README already contains corrected Docker run instructions and no longer contains the dcoker typo from the old version. This documentation fix is superseded, so we are closing the pull request.

PR #106, opened 2023-06-30
Fix(datasource): Solve the problem that the test connection takes too long when the MySQL data source tests the connection, and the ip or port input is wrong.

happyJonychen:
When adding or editing MySQL data sources, if there is an error in ip or port, it will take a long time to test the connection. You can set a timeout to shorten this time.


fuchanghai:
hi @happyJonychen  我觉得可以把这些配置拿出来作为配置项，WDYT? cc @zhuangjiaju

PR #108, opened 2023-07-01
feat: hive dataSource

ytwp:
1. Improve the HIVE data source

2. Fix the buildColumn method, which will cause an error when the field does not exist.


openai0229:
Thank you for the Hive contribution. Current Community main now includes a complete Hive datasource implementation with frontend connection configuration, plugin registration, metadata and database management, SQL building, and SQL parsing. The original datasource addition is therefore superseded by the current architecture and this pull request will be closed. The old defensive ResultSet metadata change is not directly portable; if the same JDBC edge case is reproducible on Community 5.3.0 or later, please open a focused issue or pull request from current main with the driver version and reproduction steps.

PR #247, opened 2023-07-14
feat: Fixed an issue where selecting password for MySQL SSH connection resulted in an error: 'PrivateKey Path must be not empty!'

heishuaiguo:
Fixed an issue where selecting password for MySQL SSH connection resulted in an error: 'PrivateKey Path must be not empty!'


openai0229:
Thank you for the contribution. The current SSH implementation uses private-key authentication only when keyFile is present and otherwise creates a username and password session, which covers the original PrivateKey Path validation problem. The old implementation is superseded, so we are closing this pull request. Please report a current connection case if this still reproduces on Community 5.3.0 or later.

Issue #705, opened 2023-10-31
Bug: Error in switching database source to mysql.


Lionelpang:
### Chat2DB Version

3.0.8

### Describe the bug

Error reported after switching data source to mysql. Some error information is as follows:
Caused by: java.sql.SQLSyntaxErrorException: Table 'stumanage.DATA_SOURCE' doesn't exist
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:120)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
	at com.mysql.cj.jdbc.StatementImpl.executeInternal(StatementImpl.java:763)
	at com.mysql.cj.jdbc.StatementImpl.execute(StatementImpl.java:648)
	at com.zaxxer.hikari.pool.ProxyStatement.execute(ProxyStatement.java:94)
	at com.zaxxer.hikari.pool.HikariProxyStatement.execute(HikariProxyStatement.java)
	at org.flywaydb.core.internal.jdbc.JdbcTemplate.executeStatement(JdbcTemplate.java:201)
	at org.flywaydb.core.internal.sqlscript.ParsedSqlStatement.execute(ParsedSqlStatement.java:95)
	at org.flywaydb.core.internal.sqlscript.DefaultSqlScriptExecutor.executeStatement(DefaultSqlScriptExecutor.java:210)
	... 40 common frames omitted

Caused by: java.sql.SQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SCHEMA, DELETED, USER_ID)
VALUES (1, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP, '学�' at line 1
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:120)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
	at com.mysql.cj.jdbc.StatementImpl.executeInternal(StatementImpl.java:763)
	at com.mysql.cj.jdbc.StatementImpl.execute(StatementImpl.java:648)
	at com.zaxxer.hikari.pool.ProxyStatement.execute(ProxyStatement.java:94)
	at com.zaxxer.hikari.pool.HikariProxyStatement.execute(HikariProxyStatement.java)
[...]

PR #706, opened 2023-10-31
Fix bug: Error #705 when switching database source to mysql

Lionelpang:
Fix the problem of error reporting under mysql.


openai0229:
Thank you for the contribution. The relational JPA entity, mapper, and Flyway migration architecture involved in this MySQL initialization issue has been removed from the current Community edition and replaced by the current local storage model. The old patch is no longer applicable, so we are closing the pull request.

PR #871, opened 2023-11-21
The database type is judged by enumeration.

tmlx1990:
The database type is judged by enumeration.


openai0229:
Thank you for the contribution. Database type handling has since been consolidated around the current DatabaseTypeEnum and unified SQL builder and parser paths. The old SQLExecutor implementation changed by this pull request no longer exists, so the pull request is superseded and will be closed.

PR #906, opened 2023-12-01
When creating a new database, the sorting order selection function is added.

tmlx1990:
When creating a new database, the sorting order selection function is added to realize the database: MySQL、Sqlite、PostgreSQL。

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,7 @@ Redis and MongoDB are partially supported , Hbase、Elasticsearch、openGauss、
 
 ### CONFIGURE CUSTOM AI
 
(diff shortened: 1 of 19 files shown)


openai0229:
Thank you for the contribution and for proposing collation selection during database creation. MySQL collation selection is now implemented on main in commit f80d70c4, using backend metadata, charset-aware filtering, and the backend-provided default collation. We intentionally expose this only for MySQL: SQLite collations are column-level, while the current PostgreSQL database builder semantics need correction before that option can be enabled safely. Since the MySQL requirement is now covered, we are closing this PR. We appreciate your contribution!

PR #917, opened 2023-12-05
You can select the character set field when creating a new database.

tmlx1990:
You can select the character set field when creating a new database.
At present, MySQL has been implemented.


openai0229:
Thank you for the contribution and for highlighting this missing create-database capability. MySQL character set selection is now implemented on main in commit f80d70c4. The options are loaded from backend metadata and included in the generated SQL preview. Since this work is now covered by the current implementation, we are closing this PR. We appreciate your contribution!

PR #955, opened 2023-12-13
Fix: #874 BI report Column has no information to show.

tmlx1990:
#874 
image


openai0229:
Thank you for the contribution. The current Community chart implementation already configures bar-chart tooltips and derives selectable chart columns from the current result metadata flow. The old dashboard implementation in this pull request has been replaced, so this change is superseded and will be closed.

PR #1043, opened 2023-12-28
Reject the return of sensitive information of shared connection when obtaining the list of data sources.

DimonHo:
api/connection/datasource/list  interface，the user name and password of the shared connection in the return list are removed from the return to avoid the disclosure of sensitive information.


shanhexi:
提到dev 解决一下冲突

PR #1086, opened 2024-01-09
DATETIME and TIMESTAMP types of MySQL database are added to update attributes according to the current timestamp.

tmlx1990:
DATETIME and TIMESTAMP types of MySQL database are added to update attributes according to the current timestamp.


openai0229:
Thank you for the contribution. Current Community main persists the onUpdateCurrentTimestamp column property and generates ON UPDATE CURRENT_TIMESTAMP in the MySQL column SQL builder. This feature is already present in the current implementation, so the older pull request is superseded and will be closed.

PR #1099, opened 2024-01-11
fix: #1093 Fix the problem that modifying table field names leads to SQL error generation.

tmlx1990:
#1093 Fix SQL error caused by MySQL modifying table field names.


openai0229:
Thank you for identifying this issue and contributing a fix. We have now implemented the rename-and-reorder handling on main in commit 702b9e7f, including preserving FIRST/AFTER column positions and adding regression coverage. Since the current implementation supersedes this PR, we are closing it. We appreciate your contribution!

PR #1145, opened 2024-01-29
Modify the calling method of openai, and use function calling to reduce the size of the first input token.

hejianjun:
Updated the interface of openai, refer to.`https://cookbook.openai.com/examples/how_to_call_functions_with_chat_models` Let ChatGPT actively obtain the data structure of the table
The next step is to add an automatic option to the front-end component of the selection table, which can automatically obtain the database structure to write sql when selecting.
image


openai0229:
Thank you for the contribution. The current Community AI stack has been rebuilt on Spring AI with the current tool-calling manager and compatible streaming function-call merger, while the legacy OpenAI client changed by this pull request no longer exists. The pull request branch also contains hundreds of unrelated changes from a long-diverged fork, so it cannot be safely integrated. We are closing it as superseded. A remaining token-usage issue should be demonstrated with a focused benchmark and a new pull request from current main.

PR #1147, opened 2024-01-29
fix mysql buildCreateTableSql error

openai0229:
Fix the sql statement error of mysql generation table.


openai0229:
Thank you for the contribution. The current MySQL create-table builder has been rewritten and filters invalid columns and indexes without relying on the old trailing-comma truncation path. The original buildCreateTableSql failure path has therefore been replaced, so this pull request is superseded and will be closed.

PR #1188, opened 2024-02-28
fix: 1186 Fixed the problem that MySQL cannot obtain table comments.

tmlx1990:
#1186 Fixed the problem that MySQL cannot obtain table comments.


openai0229:
Thank you for the contribution. Current Community main reads MySQL table comments directly from TABLE_COMMENT and also handles column and index comment metadata in the current plugin. The original issue is covered by the current implementation, so this pull request is superseded and will be closed.

PR #1202, opened 2024-03-06
Add new database Snowflake

tmlx1990:
Add new database Snowflake。


openai0229:
Thank you for adding Snowflake support. Current Community main now contains a complete Snowflake plugin with metadata, manager, SQL builder, parser, and client registration. The feature has been independently integrated into the current plugin architecture, so this older pull request is superseded and will be closed.

PR #1208, opened 2024-03-07
Adaptation deployment data source: MYSQL

188:
## Modify the initialization script
- ssl Indicates `ssl`  to match the MYSQL keyword
- schema is changed to `schema`  to match the mysql keyword
- Change bigint(20) to bigint: Integer display width is deprecated and will be removed in a future release.
- MODIFY the field length. Use: Replace ALTER with modify to adapt MYSQL
## Modify the DO MYSQL keyword processing
- schema field Added @TableField(" `schema` ") annotation
- ssl field Added @TableField(" `ssl` ") annotation

## Test method
- ai.chat2db.server.domain.repository.Dbutils  initDataSource method to modify the data source for MYSQL in 8


openai0229:
Thank you for the contribution. The JPA entity and Flyway migration model changed by this pull request has been removed from the current Community edition. Community now uses the new local datasource and workspace storage architecture, with Flyway disabled in the Community runtime. This patch is no longer applicable and will be closed.

PR #1321, opened 2024-04-16
Fix the problem that hive cannot view the library.

tmlx1990:
Fix the problem that hive cannot view the library.


openai0229:
Thank you for the contribution. The current Community Hive plugin now loads Hive databases through SHOW DATABASES and exposes them through the schema layer used by the current Hive datasource model. This covers the original issue where Hive databases could not be viewed. The pull request also targets the removed server architecture and contains unrelated historical changes, so it is superseded by current main and will be closed.

PR #1331, opened 2024-04-20
Fix clickhouse tables query

openai0229:


openai0229:
Thank you for the contribution. The current ClickHouse metadata implementation queries system.tables with a parameterized database filter and supports an optional table-name filter, covering the original table-listing issue in the current plugin architecture. This older pull request is superseded and will be closed.

PR #1339, opened 2024-04-27
Db data import

openai0229:


openai0229:
Thank you for the substantial contribution. The Community import and export subsystem has since been independently rebuilt around the current controller, domain service, task, and frontend flows, with CSV, XLS, XLSX, and SQL support in the current product scope. This pull request targets the removed server architecture and cannot be integrated into current main, so we are closing it as superseded. Any missing format or workflow should be proposed as a focused change from the current main branch.

PR #1376, opened 2024-05-11
Add SUNDB

tmlx1990:
Added SUNDB database support


openai0229:
Thank you for adding SUNDB support. Current Community main now contains a complete SUNDB plugin and the corresponding client and database-type registration. The feature has been independently integrated into the current plugin architecture, so this older pull request is superseded and will be closed.

PR #1429, opened 2024-06-14
add Xugudb

tmlx1990:
add Xugudb


openai0229:
Thank you for adding XuguDB support. Current Community main now contains a complete XuguDB plugin with configuration, metadata, manager, SQL builder, parser, and client registration. The feature has been independently integrated into the current plugin architecture, so this older pull request is superseded and will be closed.

PR #1430, opened 2024-06-15
Fix pgsql table ddl

openai0229:


openai0229:
The PostgreSQL table DDL generation path has since been rebuilt in the current Community PostgreSQL plugin. This older implementation is now superseded by current main, so we are closing the pull request.

PR #1443, opened 2024-06-26
fix-sqlserver-ddl

openai0229:


openai0229:
The SQL Server DDL generation path has since been rebuilt in the current Community SQL Server plugin, including the current table DDL and column metadata handling. This older implementation is now superseded by current main, so we are closing the pull request.

Issue #1446, opened 2024-06-28
Bug: oracle 直接在表格修改数据和导出的sql有问题 时间字段未添加timestamp


pantheonyang:
### Chat2DB Version

3.2.2

### Describe the bug

oracle 导出数据生成的sql和在表格上直接修改数据生成的sql不对
![微信图片_20240628125718](https://github.com/chat2db/Chat2DB/assets/147964577/f32e18e4-a3aa-44b5-842f-c00d6eb956b6)
正确的sql应该是UPDATE ASSET.TEST SET "date" = TIMESTAMP '2024-06-28 12:55:37.000000' WHERE ID = '213';

openai0229:
Thanks for reporting this. Chat2DB Community 5.3.x is a major new baseline rather than a patch on the historical release line. The database client core, local storage, packaging, and AI integrations have changed substantially, so this report against the previous architecture cannot be treated as a current 5.3.1 defect. We are closing this historical issue as outdated, not claiming that the exact symptom was individually fixed. For Community, please back up existing data and test with a fresh 5.3.1 installation because historical Community releases do not support an in-place upgrade. For Pro or Local, please test with the current corresponding edition. If the problem is still reproducible, open a new issue using the current template with the edition, OS, database version, steps, and logs.

openai0229:
Thanks for reporting this. Chat2DB Community 5.3.x is a major new baseline rather than a patch on the historical release line. The database client core, local storage, packaging, and AI integrations have changed substantially, so this report against the previous architecture cannot be treated as a current 5.3.1 defect. We are closing this historical issue as outdated, not claiming that the exact symptom was individually fixed. For Community, please back up existing data and test with a fresh 5.3.1 installation because historical Community releases do not support an in-place upgrade. For Pro or Local, please test with the current corresponding edition. If the problem is still reproducible, open a new issue using the current template with the edition, OS, database version, steps, and logs.

openai0229:
[...]

PR #1448, opened 2024-06-28
fix  export  sql error  and the   generated sql when modify web tabledata error

pantheonyang:
#1446 Optimize the process of exporting SQL and adopt Druid for generating change SQL statements from web table modifications, ensuring compatibility across different database types.
优化导出sql和前端页面修改表格生成的变更sql采用druid生成，兼容数据库类型


openai0229:
Thank you for the contribution. DML export and table-mutation SQL generation have since been rebuilt in the current Community domain services, including the current database-type-aware export and table update paths. The files changed by this pull request belong to the removed server layout, so the pull request is superseded and will be closed. Please open a focused report against current main if a specific export or table-editing case remains broken.

Issue #1555, opened 2024-11-01
Bug: 自定义AI使用阿里云百炼的通义千问接口报错


Jiang0977:
### Chat2DB Version

3.4.1

### Describe the bug

![image](https://github.com/user-attachments/assets/817ac238-7c6b-4055-88ad-5e190fe9bf96)
上面是通义千问AI配置，但是使用是报错：Tongyi Chat AI error：{"error":{"code":"missing_required_parameter","param":"message","message":"you must provide a messages parameter","type":"invalid_request_error"}}

tmlx1990:
您好，是否是模型参数写的不对？我看官网给的案例是 "model": "qwen-plus"
https://bailian.console.aliyun.com/?spm=a2c4g.11186623.0.0.1a5e55efUuK0y2&accounttraceid=494cccbc11014cf1a6f54cba1bc77e68dsju#/model-market/detail/qwen-plus-0919
<img width="817" alt="image" src="https://github.com/user-attachments/assets/e9da018a-8886-4686-a92e-e1552aa98e60">

tmlx1990:
这块是有点问题，需要使用 如下地址调用：https://dashscope.aliyuncs.com/api/v1/services/aigc/text-generation/generation

<img width="947" alt="image" src="https://github.com/user-attachments/assets/b7863bb8-ec0f-4a29-b5f7-a64f53623df1">

onlyyou2023:
@tmlx1990 https://dashscope.aliyuncs.com/api/v1/services/aigc/text-generation/generation
按照这个地址调用通义千问,会返回一大堆非sql的文本,这块是调通义千问接口参吗,还是如何处理.
或者官方推荐哪种第三方ai调用方式是简单好用的

LelandACM:
> @tmlx1990 https://dashscope.aliyuncs.com/api/v1/services/aigc/text-generation/generation 按照这个地址调用通义千问,会返回一大堆非sql的文本,这块是调通义千问接口参吗,还是如何处理. 或者官方推荐哪种第三方ai调用方式是简单好用的

此问题解决了吗？

tmlx1990:
> > @tmlx1990 https://dashscope.aliyuncs.com/api/v1/services/aigc/text-generation/generation 按照这个地址调用通义千问,会返回一大堆非sql的文本,这块是调通义千问接口参吗,还是如何处理. 或者官方推荐哪种第三方ai调用方式是简单好用的
> 
[...]

PR #1563, opened 2024-11-09
Fix:1555修复通义千问返回结果不能解析问题

tmlx1990:
#1555 修复通义千问返回结果不能解析问题


openai0229:
Thank you for the contribution and for investigating the Tongyi Qianwen response parsing issue. The current main branch no longer uses the legacy TongyiChatAIEventSourceListener. AI streaming is now handled through Spring AI, and Qwen is supported through the OpenAI-compatible provider configuration with a custom base URL and model. Because this patch targets the removed legacy integration and also contains unrelated changes, it is no longer applicable to the current codebase. We are closing this PR, and we appreciate the original report and implementation effort.

PR #1590, opened 2024-12-05
add language for Korean

hkjang:
add language for Korean


openai0229:
Thank you @hkjang for contributing Korean localization to Chat2DB. We have now implemented Korean support independently against the current architecture on main in commit 361a0710, including the complete frontend catalog, backend and JCEF messages, Korean README, locale integration, and CI parity checks. Because the new implementation supersedes this older PR, we are closing it. We sincerely appreciate the time and care you invested in making Chat2DB more accessible to Korean-speaking users.

PR #1683, opened 2025-03-12
Bug fixed: https://github.com/CodePhiliaX/Chat2DB/issues/1677

Sunny955:


Sunny955:
Added the null checks for chat2dbProperties

openai0229:
Thank you for the contribution. The GatewayClientService login QR-code flow addressed by this pull request is no longer part of the Community edition, and the old endpoint and service have been removed. This patch is therefore no longer applicable and will be closed.

Issue #1597, opened 2024-12-14
优化: 日期格式的数据能不能改成日期控件的方式


cpyong:
优化: 日期格式的数据能不能改成日期控件的方式，目前文本框操作起来真的是很不方便，谢谢！

openai0229:
Thanks for the suggestion. Chat2DB Community 5.3.x is a major new product baseline, and the feature architecture and roadmap have changed substantially since this request was filed. We are closing this historical request as part of resetting the pre-5.3 backlog; this does not mean the feature was implemented or rejected on its merits. If this capability is still needed in 5.3.1, please open a new feature request using the current template and describe the current edition, workflow, expected behavior, and concrete use case.

PR #1735, opened 2025-04-12
feat:日期格式的数据使用日期控件提交

1149538837:
#1597


openai0229:
Thank you for the contribution. The current Community result table now registers dedicated date, time, datetime, and timestamp editors backed by the current CanvasTable editor implementation. This requirement has therefore been independently implemented and the old client patch is superseded. If a date-editing case still fails on Community 5.3.0 or later, please report that case against the current main branch.

PR #1748, opened 2025-05-16
Fixed TongYiQiqnWen AI error.

gui0923:
The new API format and response format of Qwen have changed a lot. I have modified them according to the latest format and verified it on qwen-plus.


openai0229:
Thank you for the contribution. The legacy direct Tongyi and Qwen client and response model changed by this pull request have been removed from Community. Current main uses the new model-configuration and streaming adapter architecture instead, so this patch is no longer applicable and will be closed. If a Qwen-compatible endpoint fails in the current model flow, please open a focused issue or pull request from current main.

PR #1770, opened 2025-08-22
Added Spanish Language

Adrc95:
Added Spanish translations


openai0229:
Thank you @Adrc95 for contributing Spanish localization to Chat2DB. We have now implemented Spanish support independently against the current architecture on main in commit 361a0710, including the complete frontend catalog, backend and JCEF messages, Spanish README, locale integration, and CI parity checks. Because the new implementation supersedes this older PR, we are closing it. We sincerely appreciate the time and care you invested in making Chat2DB more accessible to Spanish-speaking users.

PR #1876, opened 2026-07-20
 feat(ai): standardize AI tool results as JSON

### LINKED ISSUE #1879: Standardize AI tool results as structured JSON
### Chat2DB Edition

  Chat2DB Community

  ### User Problem

  ## Problem

  AI database tools currently return free-form text with inconsistent formats across tools.

  For example, metadata tools may return newline-separated text, schema tools may return DDL-like text,
  and SQL execution tools may return table-like text mixed with status fields. This makes it harder for
  the model and other tool consumers to reliably determine:

  - whether the tool call succeeded
  - which tool produced the result
  - what data should be consumed
  - what error occurred
  - how SQL result columns map to row values

  This is especially fragile for `execute_sql`, where the model has to infer rows and columns from a
  text/table preview.

  ## Proposed solution

  Introduce a standard JSON result contract for AI tool outputs:

  ```json
  {
    "success": true,
    "tool": "execute_sql",
    "summary": "...",
    "data": [],
    "errorCode": null
  }
  ```

  Each AI tool should return this shape consistently.

  Expected behavior:

  - `success` indicates whether the tool call succeeded.
  - `tool` identifies the tool name.
  - `summary` provides a human-readable summary.
  - `data` contains structured payloads such as datasource lists, table lists, schema information, or
  SQL result rows.
  - `errorCode` provides a stable failure reason when `success` is false.

  ## Scope

  Standardize result output for these AI database tools:

  - `list_all_datasources`
  - `list_all_databases`
  - `list_all_schemas`
  - `list_all_tables`
  - `get_tables_schema`
  - `execute_sql`

  Also ensure adapter-level failures, including MCP tool failures, return the same JSON shape.

  ## Benefits

  - Improves model reliability when consuming tool results.
  - Avoids parsing table-like or natural-language text for SQL results.
  - Provides a consistent format for both success and failure.
  - Makes AI trace rendering and future frontend display easier.
  - Makes MCP tool result handling more predictable.

  Related PR: #1876

  ### Product area

  AI

  ### Proposed outcome

  AI database tool results should use a consistent JSON structure so the model can reliably consume
  success state, error information, schema metadata, and SQL result rows from fixed fields instead of
  parsing free-form text.

haizhi03:
## Related issue

  Closes #1879

  ## Summary

  Standardize AI database tool results as structured JSON instead of free-form text.

  Previously, AI tools returned plain strings with different formats across tools. The model had to
  infer success, errors, metadata, schema details, and SQL result rows from natural language or table-
  like text.

  This PR introduces a unified `AiToolResult` shape:

  ```json
  {
    "success": true,
    "summary": "...",
    "data": [],
    "errorCode": null
  }
  ```

  The tool name is not included in the result payload because it is already available from the tool
  invocation / callback context.

  This makes tool result consumption more predictable for the model and easier to extend for trace
  rendering, MCP usage, and future frontend display.

  ## Affected surfaces

  - [ ] Frontend / Web
  - [x] Backend / API / Storage
  - [ ] CI / Build / Release
  - [ ] Documentation only

  ## Verification

  - Commands and results:

  ```bash
  mvn -B test -f chat2db-community-server/pom.xml \
    -pl chat2db-community-domain/chat2db-community-domain-core -am \
    -Dtest=AiToolServiceImplTest \
    -Dmaven.test.skip=false \
    -DskipTests=false \
    -Dsurefire.failIfNoSpecifiedTests=false
  ```

  Result: `BUILD SUCCESS`, `Tests run: 2, Failures: 0, Errors: 0, Skipped: 0`.

  ```bash
  mvn -B test -f chat2db-community-server/pom.xml \
    -pl chat2db-community-web -am \
    -Dtest=WebClientParameterFilterTest \
    -Dmaven.test.skip=false \
    -DskipTests=false \
    -Dsurefire.failIfNoSpecifiedTests=false
  ```

  Result: `BUILD SUCCESS`, `Tests run: 3, Failures: 0, Errors: 0, Skipped: 0`.

  - Manual verification: N/A
  - UI evidence: N/A

  ## Risk and compatibility

  - Public API or stored data:
    AI tool return content changes from free-form text to JSON strings. This may affect model/tool
    consumers that directly parse tool return text. No stored data format is changed.

  - Database or driver compatibility:
    N/A. This PR does not change database drivers or SQL execution behavior.

  - Network, privacy, or security:
    N/A. This PR does not change network access, credential handling, or data exposure rules.

  - Community / Local / Pro boundary:
    N/A. This PR only changes community AI tool result formatting and does not change product boundary
    behavior.

  - Backward compatibility:
    Existing tool methods still return `String`, but the string payload now follows a JSON contract.
    Prompts were updated to instruct the model to consume `success`, `summary`, `data`, and
    `errorCode`.

  ## Reviewer map

  - Start here:
    `AiToolResult` and `AiToolServiceImpl` result wrapping:
    - `chat2db-community-domain-api/src/main/java/ai/chat2db/community/domain/api/model/ai/
    AiToolResult.java`
    - `chat2db-community-domain-core/src/main/java/ai/chat2db/community/domain/core/impl/ai/
    AiToolServiceImpl.java`

  - Failure condition:
    AI tool consumers fail to handle JSON result strings, or model responses regress because prompts no
    longer match tool output shape.

  - Rollback or disable path:
    Revert this PR to restore the previous free-form string tool results.

  ## Contributor declaration

  - [x] I disclosed substantial AI assistance below, or this PR contains no substantial AI-generated
  code.


openai0229:
Do we need the `tool` field in the result payload? The tool name is already available from the tool invocation/callback context, so returning it again in every result appears redundant. I suggest keeping the contract focused on `success`, `summary`, `data`, and `errorCode`, unless there is a concrete consumer that receives only the raw payload without any invocation metadata.

haizhi03:
Thanks for the suggestion. That makes sense since the tool name is already available from the tool
  invocation / callback context.

  I removed the `tool` field from the result payload and kept the contract focused on `success`,
  `summary`, `data`, and `errorCode`.

  The tool name is still available from the invocation context / trace metadata, but it is no longer
  duplicated in each returned result payload.

openai0229:
Thanks for the update. After reviewing the current result mapping and trace flow, I found three remaining issues that should be addressed before merging:

1. **Duplicate SQL column labels are lossy.** [`rowPreview`](https://github.com/OtterMind/Chat2DB/blob/72fc8427552ae0296f8d7da2b587a41235e1edcd/chat2db-community-server/chat2db-community-domain/chat2db-community-domain-core/src/main/java/ai/chat2db/community/domain/core/impl/ai/AiToolServiceImpl.java#L618-L631) uses the column label as a `Map` key. A query such as `SELECT a.id, b.id FROM ...` produces columns `["id", "id"]`, but the row becomes `{"id": <second value>}` and silently loses the first value. Rows should remain positional arrays aligned with `columns`, or use stable unique column identifiers.

2. **The structured row data is still display-formatted and lossy.** [`normalizeCell`](https://github.com/OtterMind/Chat2DB/blob/72fc8427552ae0296f8d7da2b587a41235e1edcd/chat2db-community-server/chat2db-community-domain/chat2db-community-domain-core/src/main/java/ai/chat2db/community/domain/core/impl/ai/AiToolServiceImpl.java#L675-L683) converts SQL `NULL` to the string `"NULL"`, rewrites control characters, and silently truncates values longer than 200 characters. This makes SQL `NULL` indistinguishable from a real string containing `NULL` and means `data` no longer represents the actual query result. Structured data should preserve JSON `null` and raw values. If preview truncation is intentional, it should use explicit fields such as `truncated` and `originalLength`.

3. **The existing frontend trace will display the raw JSON string.** [`AiToolAdapter.emit`](https://github.com/OtterMind/Chat2DB/blob/72fc8427552ae0296f8d7da2b587a41235e1edcd/chat2db-community-server/chat2db-community-web/src/main/java/ai/chat2db/community/web/api/adapter/ai/AiToolAdapter.java#L140-L145) puts the serialized JSON directly into trace `content`, while the current frontend renders [`entry.content` directly inside a `<pre>`](https://github.com/OtterMind/Chat2DB/blob/72fc8427552ae0296f8d7da2b587a41235e1edcd/chat2db-community-client/src/blocks/AI/index.tsx#L1735-L1741). This PR therefore affects the frontend even though Frontend/Web is not marked as affected. The trace should receive a human-readable summary separately, or the frontend should parse and format the structured result.

Please also add behavior tests for duplicate column labels, SQL null values, truncation semantics, and trace rendering. The current tests only verify serialization of the outer envelope and do not exercise any of these transformations.

haizhi03:
Thanks for the detailed review. I updated the PR to address the remaining issues.

  - SQL result rows are now positional arrays aligned with `columns`, so duplicate column labels such
  as `["id", "id"]` no longer overwrite values.
  - Structured row values now preserve the actual query result:
    - SQL `NULL` remains JSON `null`
    - strings are kept as raw values in `data.rows`
    - display formatting is kept only in the human-readable `text` preview
  - Added explicit row preview metadata for bounded result previews:
    - `rowCount`
    - `previewRowCount`
    - `rowsTruncated`
  - Tool trace rendering no longer receives the full serialized JSON as display content. The trace
  content uses the result `summary`, while the model/tool caller still receives the full JSON payload.
  - Updated `IAiToolService` comments to document that these methods return serialized `AiToolResult`
  JSON strings instead of context text.
  - MCP `text2sql` success responses are also wrapped in the same standard JSON contract.

  I also added behavior tests for duplicate column labels, SQL null/raw value preservation, preview
  truncation metadata, trace summary rendering, and MCP `text2sql` JSON success responses.

  Verification was rerun with `-Dmaven.test.failure.ignore=false`:

  - `AiToolServiceImplTest`: 4 tests passed
  - `AiToolAdapterTest`: 2 tests passed
  - `AiToolMcpAdapterTest`: 1 test passed
  - `WebClientParameterFilterTest`: 3 tests passed

openai0229:
### Remaining design concern: keep the service result boundary typed and transport-independent

Thanks for addressing the previous data-fidelity and trace-rendering issues. The updated payload is much safer for model consumption, but I think the result boundary is still placed at the wrong layer.

#### Current flow

`IAiToolService` currently returns a serialized `AiToolResult` JSON `String`. This makes `AiToolServiceImpl` responsible for all of the following:

1. Resolving connection context and executing the database operation.
2. Applying SQL safety and execution policies.
3. Converting domain results such as `WorkspaceDataSource`, `SimpleTable`, `Schema`, and `ExecuteResponse` into AI-specific maps.
4. Creating human-readable summaries and protocol error codes.
5. Wrapping the result in `AiToolResult`.
6. Serializing the wrapper to JSON.

The caller then receives only the serialized, projected result. It cannot use the original typed execution result, and `AiToolAdapter` has to parse the JSON again just to extract `summary` for trace output.

For `execute_sql`, this is especially visible: the original `ExecuteResponse` objects are converted to a selected `Map<String, Object>` representation, limited to a 50-row preview, supplemented with a text representation, wrapped in `AiToolResult`, and serialized before leaving the service.

#### Proposed responsibility split

The service should return strongly typed business results and allow failures to propagate as typed exceptions:

```text
IAiToolService
  - resolve and bind database context
  - validate arguments and SQL execution policy
  - execute the operation
  - record operation logs
  - return typed business data
  - throw typed business exceptions
        |
        v
AiToolResultConverter / Assembler
  - select fields intended for AI consumption
  - apply row preview limits
  - construct summaries
  - create typed AI payload DTOs
        |
        v
AiToolAdapter / AiToolMcpAdapter
  - catch and map exceptions to stable error codes
  - wrap the result in AiToolResult<T>
  - emit trace information
  - serialize the final transport payload
```

A possible service contract would be:

```java
List<WorkspaceDataSource> listAllDataSources(AiToolContextRequest request);

List<SimpleTable> listAllTables(AiListTablesRequest request);

List<Database> listAllDatabases(
        Long dataSourceId,
        AiToolContextRequest request);

List<Schema> listAllSchemas(
        String databaseName,
        Long dataSourceId,
        AiToolContextRequest request);

List<ExecuteResponse> executeSql(AiExecuteSqlRequest request);

List<TableSchemaResult> getTablesSchema(
        AiGetTablesSchemaRequest request);
```

Here, "raw" means a lossless, strongly typed business result. It does not require exposing an internal driver object directly. For example, `TableSchemaResult` can be a typed use-case result containing the table name, DDL, and table metadata because that operation combines multiple underlying calls. It should not contain JSON, AI summaries, or display-formatted text.

#### Error handling

Service errors should not be converted into `AiToolResult.failure(...)` inside `AiToolServiceImpl`.

Suggested behavior:

- Blank or invalid input throws a typed invalid-argument exception.
- SQL requiring manual confirmation throws a typed confirmation-required exception.
- Datasource or metadata lookup failures throw typed query exceptions.
- SQL execution failures throw a typed execution exception.
- Empty query results return an empty collection and remain successful.
- The service still uses `finally` for connection-context cleanup and continues recording operation logs.
- Unexpected exceptions propagate to the adapter boundary.

The adapters can then map these exception types to the public tool protocol:

```java
try {
    List<ExecuteResponse> result = aiToolService.executeSql(request);
    AiToolResult<SqlToolData> output = converter.fromExecuteResult(result);
    emitTrace(toolContext, toolName, output.getSummary());
    return serializer.toJson(output);
} catch (InvalidArgumentException e) {
    return failure("INVALID_ARGUMENT", e.getMessage());
} catch (SqlConfirmationRequiredException e) {
    return failure("SQL_REQUIRES_MANUAL_CONFIRMATION", e.getMessage());
} catch (SqlExecutionException e) {
    return failure("SQL_EXECUTION_FAILED", e.getMessage());
} catch (Exception e) {
    log.error("AI tool execution failed", e);
    return failure("TOOL_CALL_FAILED", "Tool execution failed.");
}
```

This ensures that exceptions never leak directly to the model while keeping the error envelope at the transport boundary where it belongs.

#### Typed output contract

`AiToolResult` should also be generic instead of using `List<Object>`:

```java
public class AiToolResult<T> {
    private boolean success;
    private String summary;
    private T data;
    private String errorCode;
}
```

Tool-specific payloads such as `SqlToolData`, `DataSourceToolData`, and `TableSchemaToolData` should be explicit DTOs rather than `Map<String, Object>`. A shared converter and serializer can be used by both the regular AI adapter and the MCP adapter so the standardized JSON contract remains consistent without moving presentation concerns into the service.

#### Test boundaries

The tests can then follow the same separation:

- Service tests verify typed return values, thrown exceptions, context cleanup, and operation logging.
- Converter tests verify field selection, positional rows, duplicate column labels, SQL null preservation, and preview truncation.
- Adapter tests verify exception-to-error-code mapping, JSON serialization, and trace summaries.
- MCP tests verify that the same output contract is produced at the MCP boundary.

This keeps the goal of this PR, namely consistent structured tool results, while separating execution, projection, error mapping, and serialization. At minimum, JSON serialization should move out of `IAiToolService`; preferably, the AI-specific result projection and envelope should move to the adapter/converter boundary as well.

haizhi03:
Addressed.

  IAiToolService now returns typed business results instead of serialized AiToolResult JSON. AiToolServiceImpl is limited to resolving/binding context, validating inputs and SQL policy, executing the
  operation, recording operation logs, returning typed data, and throwing typed exceptions.

  AI-specific projection now lives at the adapter/converter boundary. AiToolResultConverter builds the tool DTOs, summaries, SQL preview data, truncation metadata, and table schema payloads. AiToolAdapter
  and AiToolMcpAdapter now map typed exceptions to stable error codes, wrap results in AiToolResult<T>, emit trace summaries, and serialize the final JSON payload.

  The MCP boundary now uses the same failure envelope as the regular adapter for unexpected exceptions: TOOL_CALL_FAILED / Tool execution failed., without leaking internal exception messages.

  Verification run:
  - AiToolServiceImplTest: 14 tests, 0 failures, 0 errors
  - AiToolAdapterTest, AiToolMcpAdapterTest, AiToolResultConverterTest: 20 tests, 0 failures, 0 errors
  - chat2db-community-domain-core full *Test.java suite: 90 tests, 0 failures, 0 errors

  I also ran git diff --check successfully. Web focused tests covering the changed adapter/converter/MCP surface passed; the later attempt to run the full web module suite was interrupted after dependency-
  resolution issues in offline mode, so I am not claiming full web-module verification.

openai0229:
Thanks for moving the result envelope and serialization out of `IAiToolService`. The service boundary is much clearer now.

The broader result-completeness and large-result retrieval design is tracked separately in #1942 and Discussion https://github.com/OtterMind/Chat2DB/discussions/1941, so it does not need to expand this PR.

I found four remaining issues within the current scope:

1. **The outer SQL success state can contradict an inner result set.**

   `wrapExecuteResults` skips null `ExecuteResponse` entries, while `AiToolResultConverter` converts a null entry into a result set with `success=false`. This can produce an outer `success=true` envelope containing an inner failed result. Null execution results should either be rejected by the service or filtered consistently.

2. **Typed exception messages are exposed directly as public summaries.**

   `AiToolAdapter` and `AiToolMcpAdapter` copy `AiToolException.getMessage()` into both the tool result and trace output. Several service exceptions include the original `BusinessException` message, which can expose unstable database or driver details. Please map exception types to safe, stable summaries at the adapter boundary and keep the original cause in logs only.

3. **The new AI payload models are unnecessarily nested and inconsistently named.**

   Types such as `DataSourceToolData.Item`, `TableToolData.Item`, and `SqlToolData.ResultSet` are public wire-contract types rather than implementation-only inner classes. Generic names such as `Item` also make schema generation, debugging, and code search harder.

   Please use explicit AI-prefixed top-level payload types and keep adapter-specific transport DTOs out of `domain-api` where possible. Existing domain types such as `WorkspaceDataSource` and `ExecuteResponse` do not need an AI prefix; only the new AI-specific payload types do.

   The PR description also shows `data` as an array, while the current implementation serializes wrapper objects such as `data.datasources` and `data.results`. Please either preserve the documented contract or update it explicitly with compatibility tests.

4. **The SQL payload duplicates the same preview data.**

   `SqlToolData.ResultSet` returns structured `columns` and `rows`, a dense `rowCellMetadata` matrix that is mostly null, and a `text` field that repeats the tabular preview. The prompt already tells the model to consume structured data, and trace rendering uses only `summary`, so the duplicated text increases payload size without a clear consumer.

   `AiToolResultConverter` also combines metadata projection, SQL raw-value conversion, tabular rendering, table-schema rendering, and summary generation in one 553-line class. Please remove the redundant representation and split the SQL/schema conversion responsibilities into focused converters.

Please add regression tests for the null-result consistency and sanitized failure summaries.

openai0229 (changes requested):
The existing change request is still unresolved in the current head. The outer/inner success semantics can still contradict each other, exception messages are still exposed to model/trace consumers, nested public payload types remain in `domain-api`, and the SQL payload/converter still carries the duplicated representations called out in the prior review.

Please address the existing four review items and add the requested contract tests before requesting another review. No new maintainer review is needed until the branch contains those changes.

haizhi03:
Thanks for the detailed review. I addressed the remaining tool result contract issues in the latest update:

  - reject null and failed SQL execution results before serialization, so the outer envelope cannot contradict an inner result
  - map typed tool exceptions to stable public summaries in both the regular adapter and MCP adapter, keeping original causes in logs only
  - move the new AI-specific wire payloads to top-level web response DTOs with explicit Ai-prefixed names
  - keep the documented data array contract instead of wrapper objects
  - remove the duplicated SQL text preview and replace dense rowCellMetadata with sparse cellMetadata
  - split SQL result conversion and table schema rendering into focused converters

  Regression coverage was added for null/failed result handling, sanitized failure summaries, trace summaries, MCP parity, array payload contracts, and SQL payload shape.

  Validation:
  - AiToolServiceImplTest: 15 tests, 0 failures
  - AiToolResultConverterTest, AiToolMcpAdapterTest, AiToolAdapterTest: 27 tests total, 0 failures
  - git diff --check passed

PR #1890, opened 2026-07-21
fix(cache): degrade safely when optional cache initialization fails

### LINKED ISSUE #1979: refactor(cache): make optional cache initialization degradation testable
### Chat2DB Edition

Chat2DB Community

### Task Category

Refactoring

### Problem

`CacheManage` treats its disk-backed cache as optional, but initialization failure handling is embedded in a static initializer and is not directly testable. The code needs a narrow, explicit degraded state so expected cache initialization failures do not poison the class while fatal VM errors are not swallowed.

### Scope

- Encapsulate cache-manager construction behind a testable initialization boundary.
- Degrade to an unavailable cache state for `Exception` and `LinkageError` raised while creating the optional cache.
- Keep cache-backed reads using their existing source-function fallback when unavailable.
- Make `fuzzyDelete` and `close` no-ops when no cache manager is available.
- Add focused tests for initialization failure classification and unavailable-state behavior.

### Non-goals

- Catching `OutOfMemoryError`, `StackOverflowError`, `ThreadDeath`, or other fatal VM errors.
- Changing cache size, expiry, storage path, serialization, or public data behavior.
- Claiming that historical Issue #1724 reproduces on the current release.

### Completion Criteria

- An injected `Exception` or `LinkageError` produces a disabled cache state without escaping class initialization.
- Fatal errors outside `LinkageError` remain uncaught.
- Public cache reads continue through their fallback functions while disabled.
- `fuzzyDelete` and `close` do not dereference a missing cache manager.
- Focused tests and the Community backend build pass.

### Verification

- `mvn -B -f chat2db-community-server/pom.xml -pl :chat2db-community-domain-core -am -Dmaven.test.skip=false -DskipTests=false -Dtest=CacheManageTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -B -f chat2db-community-server/pom.xml -pl :chat2db-community-domain-core -am -Dmaven.test.skip=true package`
- `git diff --check`

### Contribution

I can submit a pull request

### Submission Checklist

- [x] I searched existing Issues and Discussions for duplicates.
- [x] I selected the affected Chat2DB edition.
- [x] This proposal does not include credentials, private data, or unpublished vulnerability details.

HandSonic:
## Related issue

Closes #1979

## Summary

Makes optional disk-cache initialization a narrow, testable degradation boundary.

- initialization failures caused by `Exception` or `LinkageError` produce an explicit unavailable state
- fatal non-linkage errors such as `OutOfMemoryError`, `StackOverflowError`, and `ThreadDeath` remain visible to the runtime
- reads fall back to the existing source functions while caching is unavailable
- `fuzzyDelete` and `close` are safe no-ops without a cache manager

The public static `CacheManage` API remains unchanged. Cache path, capacity, expiry, and JSON serialization are also unchanged.

## Affected surfaces

- [ ] Frontend / Web
- [x] Backend / API / Storage
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Commands and results:
  - `mvn -pl chat2db-community-domain/chat2db-community-domain-core -am -Dtest=CacheManageTest -Dsurefire.failIfNoSpecifiedTests=false -Dmaven.test.skip=false -DskipTests=false test` - passed; 4 tests, 0 failures/errors/skips
  - `git diff --check origin/main...HEAD` - passed
- Manual verification: N/A; failure modes are injected at the cache-manager construction boundary.
- UI evidence: N/A

## Risk and compatibility

- Public API or stored data: Public static methods and on-disk cache format remain unchanged.
- Database or driver compatibility: N/A.
- Network, privacy, or security: N/A; local optional metadata cache only.
- Community / Local / Pro boundary: Community domain cache implementation only.
- Backward compatibility: Existing cache hits, refreshes, fallback functions, invalidation, and serialization are preserved.

## Reviewer map

- Start here: `CacheManage.CacheStore.initialize`, then `CacheManageTest`.
- Failure condition: Optional cache initialization poisons the class, fatal errors are swallowed, fallback reads stop running, or lifecycle methods dereference a missing manager.
- Rollback or disable path: Revert the maintainer follow-up commit to restore the original static implementation.

## Contributor declaration

AI assistance: OpenAI Codex assisted with the maintainer follow-up, regression tests, and verification; the resulting diff was reviewed locally.


openai0229:
The `fuzzyDelete()` initialization guard looks correct. However, please narrow the `catch (Throwable)` clause. Catching every `Throwable` also swallows errors such as `OutOfMemoryError`, `StackOverflowError`, and `ThreadDeath`, after which the application should not continue as if only the optional cache failed. If the goal is to handle `NoClassDefFoundError`, use a targeted clause such as `catch (Exception | LinkageError e)` or `catch (Exception | NoClassDefFoundError e)`. The original `StateTransitionException` shown in the referenced historical issue is already covered by `Exception` in the current code, so a current-version reproduction or focused test for the intended `Error` path would also help validate this change.

HandSonic:
Fixed. Changed  to , and added an  guard at the beginning of  to prevent accessing  when cache is not initialized.

HandSonic:
Fixed. Changed `catch (Throwable e)` to `catch (Exception | LinkageError e)`, and added an `if (!init) return;` guard at the beginning of `fuzzyDelete()` to prevent accessing `cacheManager` when cache is not initialized.

openai0229 (changes requested):
The narrowed `Exception | LinkageError` boundary is a better direction than catching all `Throwable`, but this still needs a focused regression test that forces cache initialization to fail and verifies the class remains usable in the intended degraded mode.

Please also coordinate with #1925: both PRs add the same `fuzzyDelete` initialization guard, so the duplicate change should live in only one focused PR. After resolving that overlap, update the branch with `main` and rerun the required checks.

openai0229 (changes requested):
Re-reviewing the current head: this should not be merged without a current reproduction and focused test.

The failure reported in historical issue #1724 was an Ehcache `StateTransitionException`, which is already covered by the `catch (Exception e)` present on current `main`. This PR additionally catches every `LinkageError` and adds a degraded-mode guard, but it does not demonstrate a current 5.3.x failure that requires that broader boundary or test initialization failure behavior.

Issue #1724 has been closed as outdated, the branch is 54 commits behind `main`, and the `fuzzyDelete` guard overlaps #1925. I recommend closing this PR. A new change should start from a current-version reproduction, define which initialization failures are safe to degrade, and include a regression test.

openai0229 (approved):
Maintainer follow-up completed. Optional cache initialization now has an explicit unavailable state, catches only Exception and LinkageError, preserves fatal Error propagation, and the fallback/lifecycle regression tests pass locally.

PR #1896, opened 2026-07-21
fix(sqlserver): preserve IDENTITY columns when copying table structure (#1691)

### LINKED ISSUE #1691: Bug: 复制表结构的时候，把id自增的约束没有复制到新表
### Chat2DB Version

v3.0.0

### Describe the bug

我是用表复制功能，但是新表没有将原始表的自增约束给带上，丢失了

HandSonic:
## Problem

SQL Server 复制表结构使用 `CREATE TABLE ... AS SELECT ... WHERE 1=0`，此语法不保留 IDENTITY 属性、主键、默认值等列级约束。复制后的表丢失自增列定义。

## Fix

改为直接查询 `sys.columns` 获取源表的列定义，构建显式的 `CREATE TABLE` 语句，包含：

- `IDENTITY(seed, increment)` 子句
- `NULL / NOT NULL` 约束
- `DEFAULT` 值
- 计算列定义（`AS ... PERSISTED`）
- `SPARSE` 属性

复制数据时使用 `INSERT INTO ... SELECT * FROM`。

## Related

Fixes #1691

### DIFF
--- a/chat2db-community-server/chat2db-community-plugins/chat2db-community-sqlserver/src/main/java/ai/chat2db/plugin/sqlserver/SqlServerDBManager.java
+++ b/chat2db-community-server/chat2db-community-plugins/chat2db-community-sqlserver/src/main/java/ai/chat2db/plugin/sqlserver/SqlServerDBManager.java
@@ -164,15 +164,66 @@ public class SqlServerDBManager extends DefaultDBManager implements IDbManager {
     public void copyTable(Connection connection, String databaseName, String schemaName, String tableName, String newTableName, boolean copyData) throws SQLException {
         String sourceTable = buildFullTableName(databaseName, schemaName, tableName);
         String targetTable = buildFullTableName(databaseName, schemaName, newTableName);
-        String sql;
-        if (copyData) {
-            sql = String.format(SQL_COPY_TABLE_DATA, targetTable, sourceTable);
-        } else {
-            sql = String.format(SQL_COPY_TABLE_STRUCTURE, targetTable, sourceTable);
-        }
 
-        log.info(" copy table sql : {}", sql);
-        DefaultSQLExecutor.getInstance().execute(connection, sql, resultSet -> null);
+        // Build CREATE TABLE with explicit column definitions to preserve IDENTITY
+        StringBuilder createSql = new StringBuilder();
+        createSql.append("CREATE TABLE ").append(targetTable).append(" (\n");
+
+        boolean[] first = {true};
+        DefaultSQLExecutor.getInstance().preExecute(connection, SELECT_TABLE_COLUMNS,
+                new String[]{schemaName, tableName}, resultSet -> {
+                    while (resultSet.next()) {
+                        String colName = resultSet.getString("COLUMN_NAME");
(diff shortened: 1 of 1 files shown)


openai0229:
Thanks for working on the SQL Server copy-table behavior. I do not think the new hand-built `CREATE TABLE` path is safe to merge in its current form.

There are several concrete correctness problems:

1. `SELECT_TABLE_COLUMNS` already returns `COLUMN_SIZE`, `COLUMN_PRECISION`, `NUMERIC_SCALE`, `COLLATION_NAME`, and `ORDINAL_POSITION`, but this implementation only appends `DATA_TYPE`. This changes definitions such as `nvarchar(200)`, `varbinary(max)`, and `decimal(19,4)` into bare `nvarchar`, `varbinary`, and `decimal`, which can silently change length/precision/scale and corrupt or reject copied data.
2. The new `copyData` path creates real `IDENTITY` and computed columns, then executes `INSERT INTO target SELECT * FROM source`. SQL Server does not allow explicit insertion into an identity column without an explicit column list plus `IDENTITY_INSERT`, and computed columns cannot be inserted into. A table containing either feature can therefore be created successfully and then fail during the data copy, leaving a partial target table.
3. The metadata query has no `ORDER BY c.column_id`, while the data copy relies on positional `SELECT *`. Column order must not depend on an unordered catalog result. Identifier construction also does not escape `]` in column names.
4. This creates a second, partial SQL Server DDL renderer in `SqlServerDBManager`. It still omits or mishandles schema details already handled elsewhere, including type parameters, column collation, PK/UQ/check constraints, indexes, foreign keys, and other special column properties. `SqlServerMetaData.tableDDL` and `SqlServerSqlBuilder.buildCreateTable` already contain more complete metadata/DDL logic; please refactor a shared source-table-to-target-table path instead of duplicating a narrower renderer here.

The linked issue #1691 has also been closed as a historical 3.0.0 report and does not provide a current 5.3.x reproduction or SQL Server version, so this PR needs a current Issue/acceptance contract.

Before merging, please add tests covering at least:

- length, `max`, precision/scale, collation, nullability, defaults, and escaped identifiers;
- identity seed/increment values that do not fit an `int`;
- computed, identity, and ordinary columns for both structure-only and `copyData=true` paths;
- column order and failure rollback/cleanup;
- preservation expectations for constraints and indexes.

The data-copy case should be exercised against SQL Server, not only asserted as a generated string, because the current `INSERT ... SELECT *` is rejected by the database for the exact column features this PR introduces.

HandSonic:
Closing this PR per review feedback. The review identified fundamental correctness issues:

1. Missing type parameters — nvarchar(200) becomes nvarchar (no length)
2. INSERT INTO identity/computed columns will fail — SQL Server rejects explicit values for IDENTITY columns without SET IDENTITY_INSERT ON, and computed columns cannot be inserted at all
3. No ORDER BY for column_id — column order in target table is nondeterministic
4. Duplicates DDL generation logic that already exists in SqlServerMetaData.tableDDL and SqlServerSqlBuilder.buildCreateTable

The reviewer correctly noted that #1691 is a historical issue, and a proper fix should leverage the existing DDL infrastructure rather than building SQL strings inline.

Thank you for the detailed review.

PR #1897, opened 2026-07-21
fix(i18n): refresh setting.ts source hashes to repair Community CI on main

solitude-mix:
Follow-up to #1889 — repairs the [`Community CI` failure on `main`](https://github.com/OtterMind/Chat2DB/actions/runs/29811212281) (`Validate localization catalogs`):

```
i18n validation failed with 2 error(s):
- es-ES/setting.ts: English source changed; update the translation and its source hash
- ko-KR/setting.ts: English source changed; update the translation and its source hash
```

### What happened
#1889 changed the Base URL hint in `en-US/setting.ts` and updated all five locale translations, but missed regenerating `chat2db-community-client/scripts/i18n-source-hashes.json`, which pins a hash of each en-US module for the managed `es-ES`/`ko-KR` translations. Since `ci.yml` triggers on `push` only, the fork PR never ran this check and it surfaced after the merge. Sorry for the red `main`.

### What changed
`scripts/i18n-source-hashes.json` regenerated with `node ./scripts/validate-i18n.cjs --write-source-hashes` — the diff is exactly the two stale `setting.ts` entries (the translations themselves were already updated in #1889).

### How I tested it
- Before: `node ./scripts/validate-i18n.cjs` reproduces the same 2 errors as the failed run.
- After: `Validated es-ES and ko-KR against 25 frontend modules, 2 properties bundles, and 5 READMEs.` — full validation passes (key sets, placeholder/tag contracts, properties bundles, and READMEs included).


openai0229:
Verified the root cause: #1889 updated `en-US/setting.ts` and the corresponding translations, but missed regenerating `chat2db-community-client/scripts/i18n-source-hashes.json`.

I reran `node ./scripts/validate-i18n.cjs --write-source-hashes` against this PR's head; the generated manifest matches this diff byte-for-byte, and `node ./scripts/validate-i18n.cjs` passes. This is the correct Community CI repair. LGTM.

openai0229:
Thanks for the contribution. The target `setting.ts` source hash is already present on the current `main` (`b5d94e85d81c95e654f89f9ecc7881b21711a38138f115a27fb79664cb68990a`). The remaining differences are stale hashes relative to newer main-branch content rather than an outstanding fix.

This PR is now obsolete, so I am closing it.

PR #1899, opened 2026-07-21
fix(spi): fix NPE in SqlUtils.buildCanEditResult when FROM clause is …

liulipengju:
## Problem
`SqlUtils.buildCanEditResult` throws an NPE when the SQL's FROM clause
is a subquery, e.g.:

```sql
SELECT * FROM (SELECT id, name FROM users) t
```

`getSQLExprTableSource()` returns `null` for `SQLSubqueryTableSource` /
`SQLValuesTableSource` (see lines 120-121), but `buildCanEditResult`
used the return value without a null check:

```java
SQLExprTableSource sqlExprTableSource = (SQLExprTableSource) getSQLExprTableSource(
        sqlSelectStatement.getSelect().getFirstQueryBlock().getFrom());
executeResult.setTableName(getMetaDataTableName(sqlExprTableSource.getCatalog(), ...));  // NPE
```

The NPE is swallowed by the surrounding `catch (Exception e)` which
sets `canEdit = false`, so queries that should be editable are silently
marked as non-editable.

## Fix
Add a null guard, mirroring the existing pattern in the sibling
`getTableName()` method (same class, lines 108-110).

## Impact
- Restores editability for queries whose FROM is a subquery / values
  table source.
- No behavior change for normal `SELECT * FROM table` or joins.
- 5 lines added, 1 line removed — minimal, focused change.

## Reproduce
Before fix: run any `SELECT * FROM (SELECT 1) t` — UI shows it as
non-editable and a stacktrace is logged as `buildCanEditResult error`.
After fix: query is editable as expected.

### DIFF
--- a/chat2db-community-server/chat2db-community-spi/src/main/java/ai/chat2db/spi/util/SqlUtils.java
+++ b/chat2db-community-server/chat2db-community-spi/src/main/java/ai/chat2db/spi/util/SqlUtils.java
@@ -78,6 +78,17 @@ public class SqlUtils {
                     if ((sqlStatement instanceof SQLSelectStatement sqlSelectStatement)) {
                         SQLExprTableSource sqlExprTableSource = (SQLExprTableSource) getSQLExprTableSource(
                                 sqlSelectStatement.getSelect().getFirstQueryBlock().getFrom());
+                        // A derived table (subquery in FROM) or other non-expr/non-join table
+                        // source cannot be mapped to a single physical table; without this guard
+                        // tableName stays null and the edit flow would emit invalid SQL such as
+                        // "UPDATE null SET ...". getSQLExprTableSource returns null for those
+                        // cases (see below), so keep the result set non-editable, mirroring the
+                        // null-check already present in getTableName(). Follows the same
+                        // setCanEdit(false)+return pattern used above for aliased/COUNT columns.
+                        if (sqlExprTableSource == null) {
+                            executeResult.setCanEdit(false);
+                            return;
+                        }
                         executeResult.setTableName(getMetaDataTableName(sqlExprTableSource.getCatalog(), sqlExprTableSource.getSchema(), sqlExprTableSource.getTableName()));
                     }
                 } else {


chat2db-community-server/chat2db-community-spi/src/main/java/ai/chat2db/spi/util/SqlUtils.java
```diff
@@ -78,7 +78,11 @@ public static void buildCanEditResult(String sql, DbType dbType, ExecuteResponse
                     if ((sqlStatement instanceof SQLSelectStatement sqlSelectStatement)) {
                         SQLExprTableSource sqlExprTableSource = (SQLExprTableSource) getSQLExprTableSource(
                                 sqlSelectStatement.getSelect().getFirstQueryBlock().getFrom());
-                        executeResult.setTableName(getMetaDataTableName(sqlExprTableSource.getCatalog(), sqlExprTableSource.getSchema(), sqlExprTableSource.getTableName()));
+                        // 当 FROM 子句为子查询等非 SQLExprTableSource/SQLJoinTableSource 类型时，
+                        // getSQLExprTableSource 返回 null，需判空避免 NPE（与 getTableName 方法保持一致）
+                        if (sqlExprTableSource != null) {
```
openai0229: This null check prevents the NPE, but `canEdit` has already been set to `true`. When the table source cannot be resolved, `tableName` remains null and the result-set editing flow forwards it to the SQL builder, which can produce invalid SQL such as `UPDATE null SET ...`. A derived table should remain non-editable unless a physical table can be resolved. Please set `canEdit` to `false` and return when `sqlExprTableSource` is null, or move `setCanEdit(true)` after the table name is resolved. Also, please rewrite the added code comment in English to follow the codebase convention.

liulipengju:
Good catch — thanks. A plain null guard would have left canEdit=true
with tableName=null, forwarding invalid SQL to the builder.

Updated: when sqlExprTableSource is null, the result set is now kept
non-editable via setCanEdit(false)+return, matching the same pattern
already used earlier in this method for aliased columns and COUNT()
columns (lines 64-65, 70-71), and consistent with the null-check in
getTableName() (lines 155-157). The added comment is now in English.

A derived table genuinely cannot be mapped to a single physical
table, so non-editable is the correct semantic, not just a fallback.

PR #1918, opened 2026-07-21
fix(sqlserver): preserve IDENTITY columns when copying table (#1691)

HandSonic:
## Summary

Preserve SQL Server `IDENTITY` semantics when copying a table while keeping the generated DDL and data-copy SQL valid for real-world schemas.

## Changes

- Create the target table from `tableDDL()` and rewrite the target schema/table references in table, index, extended-property, and self-referencing foreign-key DDL.
- Remove explicit constraint names that would collide with the source table.
- Split SQL Server `GO` batches without treating strings, comments, or quoted identifiers as separators, including the legacy `go<TAB>exec sp_addextendedproperty` form.
- Copy data with the same explicit source/target column list, excluding computed, `TIMESTAMP`, and `ROWVERSION` columns.
- Enable `IDENTITY_INSERT` only for tables that actually contain an identity column, and always attempt to turn it off after a successful `ON`, preserving cleanup failures as suppressed exceptions.
- Preserve non-default identity seed/increment values with `BigDecimal`, normalize a single bracket-quoted table identifier at the copy boundary, and avoid invalid JDBC size metadata for `TIMESTAMP`/`ROWVERSION`.
- Return referenced schema and table names separately when rebuilding foreign keys; quote and escape constraint/column identifiers; preserve independent `ON DELETE` and `ON UPDATE` actions in SQL Server grammar order.
- Add focused regression coverage for DDL rewriting, batch splitting, column selection, identity handling, identifier quoting, referential actions, and failure cleanup.

## Validation

- Tools: 26 tests passed.
- SPI: 54 tests passed.
- SQL Server plugin: 37 tests passed.
- `git diff --check` passed.

Related to #1691

### DIFF
--- a/chat2db-community-server/chat2db-community-plugins/chat2db-community-sqlserver/src/main/java/ai/chat2db/plugin/sqlserver/SqlServerDBManager.java
+++ b/chat2db-community-server/chat2db-community-plugins/chat2db-community-sqlserver/src/main/java/ai/chat2db/plugin/sqlserver/SqlServerDBManager.java
@@ -17,6 +17,7 @@ import java.sql.PreparedStatement;
 import java.util.Date;
 
 import static ai.chat2db.plugin.sqlserver.constant.SQLConstant.*;
+import static ai.chat2db.plugin.sqlserver.constant.SqlServerDBManagerConstants.*;
 import static cn.hutool.core.date.DatePattern.NORM_DATETIME_PATTERN;
 
 import static ai.chat2db.plugin.sqlserver.constant.SqlServerDBManagerConstants.*;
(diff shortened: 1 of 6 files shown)


openai0229 (changes requested):
The current implementation is not safe for the SQL Server table shapes claimed by this PR.

- `SET IDENTITY_INSERT ... ON` is executed unconditionally and fails for tables without an identity column.
- `INSERT INTO target SELECT * FROM source` is invalid when computed columns or rowversion/timestamp columns are present; the copy needs explicit compatible column lists.
- Global `ddl.replace("[source]", "[target]")` can rewrite references outside the object name and does not safely regenerate constraint/index names.

Please redesign this around parsed metadata and explicit column/object lists, then add SQL Server integration tests for tables with no identity, identity, computed columns, rowversion, defaults, constraints, and indexes. The PR should not claim those objects are preserved until the tests prove it.

HandSonic:
Redesigned the copyTable implementation per review feedback:

**1. Conditional IDENTITY_INSERT** — no longer executed unconditionally. Now queries `sys.columns` via `SELECT_TABLE_COLUMNS` to check if the table has any identity column. Only executes `SET IDENTITY_INSERT ON/OFF` when identity columns exist.

**2. Explicit column list** — instead of `INSERT INTO target SELECT * FROM source`, now builds an explicit column list by querying metadata and filtering out:
- Computed columns (have `COMPUTED_DEFINITION`)
- timestamp/rowversion columns (auto-generated, not copyable)

Uses `INSERT INTO target (col1, col2, ...) SELECT col1, col2, ... FROM source`.

**3. Safe table name replacement** — instead of global `ddl.replace("[source]", "[target]")`, now uses `replaceFirst` with a regex targeting only the `CREATE TABLE [tableName]` line, so constraint/index references inside the DDL are not affected.

**4. Rebased on current main.**

openai0229 (approved):
Maintainer follow-up now covers explicit column selection, identity ownership and cleanup, SQL-aware DDL/batch rewriting, identifier handling, and focused regression tests. The previous blocking concerns are resolved.

openai0229 (approved):
Maintainer follow-up complete: verified identifier normalization, constraint quoting, referential-action preservation, and SQL Server grammar ordering. Tools (26), SPI (54), and SQL Server (37) tests pass.

PR #1925, opened 2026-07-22
fix(cache): init guard, corrupted entry eviction, and lock improvement

HandSonic:
## Summary

Fixes 3 verified bugs in the cache layer.

### Fixes included:
- **CacheManage.fuzzyDelete()**: added `if (!init) return;` guard to prevent NPE when static initialization failed
- **CacheManage.getList()**: evicts corrupted cache entries on deserialization failure instead of letting them persist until TTL
- **MemoryCacheManage**: replaced `String.intern()` locking with `ConcurrentHashMap` to prevent JVM string pool pollution


openai0229 (changes requested):
The corrupted-entry eviction and initialization guard look reasonable, but the lock change still introduces unbounded permanent retention. `LOCK_MAP.computeIfAbsent(lockKey, ...)` adds one entry per distinct key and no path ever removes it, so this replaces string-intern retention with another process-lifetime map leak.

Please use a bounded striped-lock implementation, or implement lifecycle-safe removal that cannot remove a lock while another thread is waiting on or using it. Add a concurrency test that covers many distinct keys and same-key mutual exclusion. Also coordinate the duplicated `fuzzyDelete` guard with #1890 so only one focused change remains.

openai0229 (approved):
Reviewed and updated. The superseded CacheManage changes were dropped in favor of current main. The remaining lock fix now uses a bounded 1,024-stripe Guava lock set instead of an unbounded map, with tests covering many distinct keys and same-key mutual exclusion. Targeted domain-core tests pass locally.

PR #1926, opened 2026-07-22
fix(plugins): view quoting, SQL injection, pagination, and misc SQL fixes

HandSonic:
## Summary

Fixes 6 verified bugs across PostgreSQL, SQL Server, and Oracle database plugins.

### Fixes included:
- **PostgreSQL view quoting**: backticks → double quotes (PG syntax)
- **PostgreSQL view comment**: added DOT separator between schema and view name
- **SQL Server view quoting**: backticks → square brackets (SQL Server syntax)
- **Oracle view comment**: added DOT separator between schema and view name
- **Oracle SQL injection**: escapes single quotes in table name via `replace("'", "''")`
- **Oracle pagination**: always applies outer SELECT wrapper to prevent CAHT2DB_AUTO_ROW_ID column leak on first page
- **SQL Server misc**: fixes TIMESTAMP dead code (`Arrays.asList()` → `Arrays.asList("TIMESTAMP")`), GO separator trailing newline, rename script bracket quoting

### DIFF
--- a/chat2db-community-server/chat2db-community-plugins/chat2db-community-oracle/src/main/java/ai/chat2db/plugin/oracle/OracleMetaData.java
+++ b/chat2db-community-server/chat2db-community-plugins/chat2db-community-oracle/src/main/java/ai/chat2db/plugin/oracle/OracleMetaData.java
@@ -137,7 +137,7 @@ public class OracleMetaData extends DefaultMetaService implements IDbMetaData {
     public List<Table> tables(Connection connection, String databaseName, String schemaName, String tableName) {
         String sql = String.format(SELECT_TABLE_SQL, schemaName);
         if (StringUtils.isNotBlank(tableName)) {
-            sql = sql + " and A.TABLE_NAME = '" + tableName + "'";
+            sql = sql + " and A.TABLE_NAME = '" + tableName.replace("'", "''") + "'";
         }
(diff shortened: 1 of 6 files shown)


openai0229 (changes requested):
This PR combines six independent behavior changes across Oracle, PostgreSQL, and SQL Server without focused dialect tests. Please split it by dialect/behavior so each fix has a clear contract and regression coverage.

At minimum, the tests need to prove Oracle pagination with both bounds after the new wrapping behavior, real SQL Server `sp_rename` quoting syntax, SQL Server TIMESTAMP/rowversion DDL behavior, and view/comment quoting for each affected dialect. After splitting and testing, please update each branch with the current `main`.

openai0229:
Rechecked against current `main`. Two concrete blockers remain: `Arrays.asList("TIMESTAMP").contains(type)` compares a `String` list with a `SqlServerColumnTypeEnum`, so the supposedly enabled TIMESTAMP branch is still unreachable; and Oracle pagination now unconditionally appends both wrappers and both bounds, including zero-bound cases, without tests proving the generated SQL. Together with the unverified `sp_rename` syntax and cross-dialect quoting changes, this is not a small safe patch to merge as one unit. Please split by dialect/behavior and add the focused SQL assertions requested in the existing review.

PR #1927, opened 2026-07-22
fix(ui): XSS, state mutation, and miscellaneous frontend fixes

HandSonic:
## Summary

Fixes 7 verified frontend bugs.

### Fixes included:
- **XSS in tree search**: escapes HTML entities before `dangerouslySetInnerHTML` to prevent stored XSS via database object names
- **Chat store state mutation**: all actions now use spread operators for immutable updates instead of mutating state in-place, fixing stale renders
- **errorLevel race condition**: uses local `effectiveErrorLevel` variable instead of mutating shared closure variable
- **useSqlExecutor onlyOne**: returns `[data[0]]` (array) instead of `data[0]` (single object), matching declared return type
- **AIChatInput useEffect**: adds `mainPageActiveTab` to dependency array so cascader data updates when switching pages
- **SSE parsing**: skips lines without colon separator instead of throwing, handling malformed SSE events gracefully


openai0229 (changes requested):
There are two blocking regressions in the current patch:

1. `updateAnswer` replaces `answers` with a one-element array (`answers: [{ ...answer, ...params }]`), which drops every existing answer after index 0. Please update the target item while preserving the rest of the array, and add a multi-answer regression test.
2. The highlight path still passes raw user input to `new RegExp(userSearch, 'gi')`. Inputs such as `[` can throw, and metacharacters such as `.` or `*` change the match semantics. Please render matched text as React text nodes without regex/HTML injection, or escape regex metacharacters and add tests for these inputs.

Please split unrelated fixes if needed so each behavior can be reviewed and tested independently.

openai0229 (approved):
Reviewed and updated. updateAnswer now preserves all answers after the first, and tree-title highlighting renders React text nodes using escaped literal matching instead of HTML injection. Added regression tests for multiple answers and literal [, ., *, case-insensitive matches, and HTML-shaped names. Targeted tests, ESLint, and the Community frontend build pass locally.

PR #1928, opened 2026-07-22
fix(ci): update i18n source hashes after en-US setting.ts change

HandSonic:
## Summary

Commit 1e99ff82 changed the `baseUrlTip` text in `en-US/setting.ts` but did not update `i18n-source-hashes.json`. This causes all CI runs to fail with:

```
es-ES/setting.ts: English source changed; update the translation and its source hash
ko-KR/setting.ts: English source changed; update the translation and its source hash
```

The es-ES and ko-KR translations were already updated in the same commit, so only the hash file needs updating.


openai0229:
Thanks for the contribution. The target `setting.ts` source hash is already present on the current `main` (`b5d94e85d81c95e654f89f9ecc7881b21711a38138f115a27fb79664cb68990a`). The remaining `common.ts` hash differences are only stale baseline differences after later main-branch changes.

There is no longer a valid net fix to merge from this branch, so I am closing this PR as obsolete.

PR #1930, opened 2026-07-22
fix(workspace): keep Data Browser active after creating a query

yashoncode:


openai0229:
Thanks for working on this. The workspace behavior has now been implemented and merged in #1975 with the finalized rule: Files/Data Browser mode changes only through an explicit user action, while console/tab changes only locate the active item inside the current mode.

This branch now conflicts with `main` and is superseded by #1975, so I am closing it to avoid maintaining two competing implementations.

PR #1966, opened 2026-07-23
Fix count SQL semicolon trimming

Aias00:
## Summary
- trim generated count SQL using its own length instead of the original SQL length
- reuse the same trailing-semicolon helper for fallback count SQL wrapping
- add focused SqlUtils tests for count SQL semicolon handling

## Verification
- mvn -B -f chat2db-community-server/pom.xml -pl :chat2db-community-spi -am -Dmaven.test.skip=false -DskipTests=false -Dtest=SqlUtilsTest -Dsurefire.failIfNoSpecifiedTests=false -Dmaven.test.failure.ignore=false test
- mvn -B -f chat2db-community-server/pom.xml -pl :chat2db-community-spi -am -Dmaven.test.skip=true package


Aias00:
close: #1970

PR #2025, opened 2026-07-24
fix(result-set): bound and clean up filter/sort render polls

### LINKED ISSUE #2023: Filter/sort pollers leak setInterval with no unmount cleanup or max-iteration guard
### Chat2DB Edition

  Chat2DB Community

  ### Chat2DB Version

  5.3.0

  ### Deployment

  Web

  ### Operating System

  macOS

  ### Operating System Version

  12

  ### Database and Version

  _No response_

  ### Problem Summary

  In `useFilterAndSort`, the filter/sort handlers poll VTable with `setInterval(..., 100)` until the rendered row-number list changes, then `clearInterval`. If the list never changes (empty result set, filter that
  yields the same rows, or the component unmounts mid-poll), the interval runs forever. The `timerId` is a local variable inside the handler closure — there is no ref and no cleanup on unmount — so after unmount
  the interval keeps firing and calls `setState`/table APIs on an unmounted component, leaking memory.

  The same pattern appears at four sites: lines 187, 219, 250, 309.

  ### Steps to Reproduce

  1. Open a result set table.
  2. Apply a filter or sort that yields the same rows as before (or trigger a filter then immediately navigate away mid-poll).
  3. Observe the `setInterval` continues to fire after unmount (unmounted-setState warnings / memory growth).

  ### Expected Behavior

  Pollers should be tracked in a ref and cleared on unmount, and should have a max-iteration / timeout guard so they cannot run forever when the row-number list never changes.

  ### Actual Behavior

  `blocks/SearchResult/components/ResultSetTable/hooks/useFilterAndSort.tsx:187-193` (and `:219`, `:250`, `:309`):

  ```ts
  const timerId = setInterval(() => {
    const afterGetRowNumberList = getRowNumberList(tableInstance);
    if (beforeGetRowNumberList.join(',') !== afterGetRowNumberList.join(',')) {
      sortAfter?.();
      clearInterval(timerId);
    }
  }, 100);
  ```

  `timerId` is local to the handler; there is no unmount cleanup and no max-iteration guard.

  ### Logs

  Fix: store timer ids in a ref, `clearInterval` them in the hook's unmount effect, and add a max-iterations cap.

  ### Impact

  Limited or cosmetic impact

  ### Workaround

  N/A — leaked intervals persist until page reload.

  ### Submission Checklist

Aias00:
## Related issue

Closes #2023

## Summary

`useFilterAndSort` polled VTable with `setInterval(..., 100)` until the
rendered `CHAT2DB_ROW_NUMBER` list changed, then `clearInterval`. If the
list never changed (empty result set, a filter yielding the same rows, or
the component unmounting mid-poll), the interval ran forever. The
`timerId` was a local variable inside each handler closure — there was no
ref and no unmount cleanup — so after unmount the interval kept firing and
called `setState`/table APIs on an unmounted component, leaking memory.
The same pattern appeared at four sites (lines 187/219/250/309).

Changes:
- Add a `pollTimerIdsRef` that tracks active poll timers.
- Add a `pollUntilRowNumbersChange(before, onSettled)` helper with a
  `MAX_POLL_ITERATIONS` cap (100 iterations x 100ms = 10s) so a poll
  cannot run forever when the row list never changes; each timer is
  registered and unregistered on settle/cap.
- Add an unmount `useEffect` that clears every active timer.
- Route all four inline poll sites (sort, filter clear-all, filter apply,
  clearAllFilters) through the helper.

The cap value (`MAX_POLL_ITERATIONS = 100`) is chosen generously: the
polls run every 100ms and VTable render settles in milliseconds, so 10s
is far above the normal settle time while still bounding the leak. It is
a single named constant and trivial to tune.

## Affected surfaces

- [x] Frontend / Web
- [ ] Backend / API / Storage
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Commands and results:
  - `npx tsc --noEmit -p chat2db-community-client/tsconfig.json` — no errors reported for `useFilterAndSort.tsx`.
  - `npx eslint src/blocks/SearchResult/components/ResultSetTable/hooks/useFilterAndSort.tsx` — no errors.
  - Confirmed only one `setInterval` remains (the helper's) and all four call sites now call `pollUntilRowNumbersChange` (lines 224/250/275/328).
- Manual verification: On an empty result set or a same-row filter, a poll now stops after the cap instead of running forever; navigating away mid-poll clears active timers via the unmount effect.
- UI evidence: N/A

## Risk and compatibility

- Public API or stored data: N/A.
- Database or driver compatibility: N/A.
- Network, privacy, or security: N/A.
- Community / Local / Pro boundary: N/A.
- Backward compatibility: Polling behavior is unchanged in the normal case (settles, calls `sortAfter`/`filterAfter`, clears the timer). Only the previously-leaking paths are bounded and cleaned up.

## Reviewer map

- Start here: `useFilterAndSort.tsx` — new `MAX_POLL_ITERATIONS` constant (line 33), `pollTimerIdsRef` + `pollUntilRowNumbersChange` helper + unmount `useEffect` (lines 52-83), and the four call sites (224/250/275/328) replacing the inline `setInterval` blocks.
- Failure condition: Intervals still leak on empty/same-row results or after unmount.
- Rollback or disable path: Revert this single commit; behavior returns to the original (leaking) polls.

## Contributor declaration


Aias00:
@openai0229 hi, are you real people? can you add my wechat: aias00

openai0229 (approved):
Reviewed all four poll sites and the timer lifecycle. The bounded polling and Set-based cleanup fix the leak; I also added cleanup when tableInstance is replaced (commit 039924f6) so old-instance timers cannot survive without an unmount. Focused ESLint, the combined Community build, and updated required checks pass. No blocking issues remain.

PR #2052, opened 2026-07-24
fix(oracle): SQL injection, view comment DOT, and pagination column leak

HandSonic:
## Summary

Fixes Oracle SQL generation in three areas:

1. **Metadata filtering safety** (`OracleMetaData.tables()`): escapes both schema and table-name SQL literals before interpolation.
2. **View DDL and comments** (`OracleSqlBuilder.buildCreateView()`): quotes schema/view identifiers with Oracle double-quote escaping, emits the missing `.` separator, handles schema-less views, and escapes comment string literals without treating backslashes as Oracle escapes.
3. **Pagination consistency** (`OracleSqlBuilder.buildPageLimit()`): always applies the outer projection so the synthetic `CAHT2DB_AUTO_ROW_ID` column does not leak on the first page, while retaining the lower-bound wrapper for later pages.

Maintainer follow-up adds focused regression coverage for quote, backslash, and apostrophe edge cases.

Split from #1926 per review feedback.


openai0229 (approved):
Reviewed the full change and added maintainer follow-up commits for Oracle pagination consistency, schema/comment SQL escaping, quoted identifiers, and regression coverage. Local reactor verification passed: tools 26, SPI 54, Oracle 15; git diff --check also passed.

PR #2053, opened 2026-07-24
fix(sqlserver): view quoting, TIMESTAMP dead code, GO separator, and rename quoting

HandSonic:
## Summary

Fixes SQL Server DDL generation in four areas:

1. **View DDL and comments** (`SqlServerSqlBuilder.buildCreateView()`): uses square-bracket identifier quoting, escapes embedded `]`, escapes comment string literals, and avoids invalid schema-level extended properties when no schema is present.
2. **TIMESTAMP/ROWVERSION DDL** (`SqlServerColumnTypeEnum.buildDataType()`): does not emit JDBC-reported precision for fixed-size row-version types.
3. **Batch separators**: preserves a trailing newline after generated `GO` separators so adjacent statements do not become one token.
4. **Column rename/alter SQL**: quotes every schema, table, constraint, and column identifier part, and escapes stored-procedure/comment string arguments.

Maintainer follow-up adds exact SQL regression tests for reserved characters and comment text.

Split from #1926 per review feedback.


openai0229 (approved):
Reviewed the full change and added maintainer follow-up commits for SQL Server identifier/comment escaping, TIMESTAMP/ROWVERSION DDL handling, GO separators, rename SQL, and regression coverage. Local reactor verification passed: tools 26, SPI 54, SQL Server 20; git diff --check also passed.

PR #2108, opened 2026-07-25
feat: universal JDBC support — add 10 databases through configuration only

Chat2DB-Pro:
## Summary

- Introduces a configuration-driven path for adding new JDBC databases to the generic plugin — no code changes required per database.
- Adds 10 new databases through configuration only: **Firebird, HSQLDB, Apache Derby, QuestDB, CrateDB, YugabyteDB, Greenplum, Trino, TimescaleDB, Apache IoTDB**.
- Routes SQL syntax plugins through a ServiceLoader-based registry (`SqlSyntaxPluginRegistry`) and adds configuration-driven identifier quoting (`ConfigurableSQLIdentifierProcessor`).
- Serves the supported-database inventory from the plugin registry via a new endpoint (`DbSupportedDatabaseController`), layered through a domain service.
- Client renders backend-registered databases dynamically without a rebuild, including per-database vector icons, host/port auto-fill, and built-in form parity.
- Fixes Derby driver classloading so all three Derby jars load through the driver classloader.
- Updates all READMEs (EN/CN/JA/KO/ES) to list the newly supported databases.

## Testing

- Unit tests for the syntax plugin registry, identifier processor, supported-database registry, and config SQL templates.
- Integration tests covering full business flows for config-only databases (Firebird, PG-wire family, Trino, embedded engines, time-series) against live containers, with scripts to start/stop the test database matrix (`script/test/start-test-databases.sh`).
- Controller-level API flow tests for config-only databases.

https://claude.ai/code/session_019y5FhjG5HB2ZLGfHu9G7gh


openai0229 (changes requested):
Thank you for the substantial work on configuration-driven JDBC support. The ServiceLoader direction, the backend inventory endpoint, and the focused registry tests are useful foundations. I found several blockers in the current head that need to be resolved before this can be merged.

1. **The PR regresses transaction ownership and connection recovery in `DefaultSQLExecutor`.**

   The current diff removes the guarded `manageTransaction` flow, rollback handling, auto-commit restoration, connection discard, and ResultSet cleanup that are already on `main`. The replacement unconditionally calls `setAutoCommit(false)`, commits, and then sets auto-commit to `true`. If the caller already owns a transaction, this commits work outside this method. If query/consumer/commit fails, it neither rolls back nor restores/discards the connection, so a pooled connection can be returned in a broken transactional state. Another changed block also removes a ResultSet try-with-resources.

   Please rebase onto the latest `main`, preserve the existing transaction/rollback/discard implementation, and apply only the required `getQualifiedTableName(...)` change on top. Add regression coverage for caller-owned transactions and consumer/query failures.

2. **Backend-provided SVG and database type data are inserted through `innerHTML` without validation or sanitization.**

   `buildIconSymbol` retains arbitrary SVG child markup and interpolates the unvalidated `dbType` into an attribute, then `init.tsx` assigns the result to `host.innerHTML`. A plugin/config entry can therefore introduce event attributes, external references, script-capable SVG content, or attribute-breaking database type text into the same-origin application.

   Prefer a fixed icon key/fallback rather than transporting raw SVG. If custom SVG is required, validate `dbType` against a strict identifier grammar and parse/sanitize the SVG with an explicit allowlist of elements and attributes; reject event attributes, scripts, foreign content, and external URLs. Do not assemble executable markup with string interpolation.

3. **The branch contains unrelated regressions and generated artifacts.**

   `chat2db-community-client/package.json` removes the existing `prebuild:web:community` guard and multiple test scripts (`test:chat-answer-update`, result inspector/Markdown, saved console, verification countdown, tree highlight, and part of SQL clipboard coverage). The PR also commits an 8.4 MB `chat2db-community-client/dist.zip` and `node_modules/.vite/.../results.json`.

   Please restore the current `main` package scripts exactly, add only the new dynamic-registry test command, remove both generated files from Git history, and extend `.gitignore` for these artifacts where appropriate.

4. **Green CI would not currently prove most of the advertised database support.**

   The Firebird, QuestDB, CrateDB, YugabyteDB, Greenplum, Trino, TimescaleDB, and IoTDB tests use JUnit assumptions and skip when their containers are absent, while the repository workflows do not start `script/test/start-test-databases.sh`. A successful regular test run can therefore report no failures while executing none of those live flows.

   Please add a CI job/matrix that starts the supported test services and reports executed versus skipped integration tests, or narrow the first PR and README support claims to the databases that are actually exercised in required CI. IoTDB should not be advertised as general database-browser support while its JDBC metadata path is explicitly outside the tested contract.

Given the 54-file, 10-database scope, please also link the repository Issue that defines the product/support contract and split the work into a reviewable foundation PR plus smaller database-family enablement PRs. That will let us validate the architecture once and verify each support claim independently.

Chat2DB-Pro:
Superseded by #2633, which rebases this work onto current `main` (this branch was 301 commits behind) and reconciles the identifier-quoting overlap with #2199.

Issue #1914, opened 2026-07-21
Security: 34 SQL Injection Vulnerabilities in Database Plugins


waterWang:
## Summary

34 SQL injection vulnerabilities found across 11 database plugin implementations in Chat2DB Community Edition.

All instances use `String.format()` to build SQL queries instead of parameterized `PreparedStatement`, allowing injection through user-controlled table names and column names.

## Affected Plugins
MySQL, Oracle, SQL Server, ClickHouse, DB2, 达梦(DM), H2, SQLite, Sundb, XuguDB

## Impact
- Data exfiltration via UNION SELECT
- Authentication bypass
- Database enumeration
- Potential RCE via xp_cmdshell (SQL Server) or UDF (MySQL)

## Fix
Replace `String.format()` with `PreparedStatement` parameterized queries and validate table names against an allowlist regex.

openai0229:
Thanks for the detailed report. Would you be willing to help us address these issues by refactoring the affected code and submitting a PR?

Since table and column identifiers generally cannot be bound through `PreparedStatement` parameters, the fix may require a combination of strict identifier validation, database-specific identifier quoting, and parameter binding for actual values. Given the number of affected plugins, the work can also be split into smaller, focused PRs if that is easier.

We would appreciate your help with the implementation.

openai0229:
Triage result: `Backlog`, P1, and removed from the public contribution queue. Identifier positions generally cannot be parameterized with `PreparedStatement`, and a database client intentionally executes user-authored SQL, so each claimed vulnerability needs a concrete untrusted source, generated statement, and privilege-impact path. Please send unpublished exploit details through `SECURITY.md`; public remediation work will be split only after validation.

PR #2176, opened 2026-07-26
fix(sundb): escape SQL identifiers and literals in metadata/DDL paths (#1914)

HandSonic:
## Related issue

Closes #1914

## Summary

Hardens SUNDB SQL generation across metadata, DDL, and inherited builder paths.

- Adds a SUNDB identifier processor with conditional SPI quoting, forced DDL quoting, embedded-quote round trips, reserved-word handling, uppercase-folding semantics, `Locale.ROOT`, and SQL string escaping.
- Routes schema, table, column, index, database, select/count/drop/truncate/update, comment, and metadata DDL identifiers through the processor.
- Replaces regex-only DEFAULT handling with a structural scanner that preserves legal literals, nested functions, casts, operators, and sequence expressions while rejecting comments, semicolons, top-level commas, unbalanced delimiters, and trailing column clauses.
- Validates unknown type expressions plus non-escapable constraint type, index direction, NULL ordering, and VARCHAR unit fragments.
- Removes obsolete manual quote-fragment constants and covers the behavior with focused SUNDB tests.

## Affected surfaces

- [ ] Frontend / Web
- [ ] Backend / API / Storage
- [x] Database plugin / Driver
- [ ] JCEF / Desktop packaging
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Commands and results: `mvn -pl chat2db-community-plugins/chat2db-community-sundb -am clean test -Dmaven.test.skip=false -DskipTests=false` from `chat2db-community-server` passed all 9 reactor modules. Tools: 26, SPI: 90, Oracle: 44, SUNDB: 25; total: 185 tests, 0 failures, 0 errors, 0 skipped.
- Manual verification: Reviewed generated SQL assertions for identifier round trips, metadata DDL, inherited builder paths, legal DEFAULT expressions, injection-shaped DEFAULT fragments, unknown types, constraint kinds, sort direction, and NULL ordering. No live SUNDB instance was available.
- UI evidence: N/A

## Risk and compatibility

- Public API or stored data: No public API or stored-data format change. Implements the shared `ISQLIdentifierProcessor.quoteIdentifierAlways` contract merged in #2234.
- Database or driver compatibility: Preserves SUNDB uppercase folding and reserved-word semantics. Existing valid DEFAULT functions and nested expressions remain supported. Invalid structural SQL fragments now fail fast with `IllegalArgumentException`.
[...]

PR #2172, opened 2026-07-26
fix(mysql): escape SQL identifiers and literals in metadata/DDL paths (#1914)

HandSonic:
## Related issue

Part of the SQL-injection hardening tracked in #1914. This PR covers the MySQL plugin and now builds on the shared forced-quote SPI merged in #2234.

## Summary

- Uses `ISQLIdentifierProcessor.quoteIdentifierAlways` for MySQL DDL and metadata identifier positions, including embedded-backtick escaping and raw-value round trips.
- Escapes MySQL string literals and validates structural tokens such as engine, charset, collation, numeric defaults, bit/hex literals, view options, index direction, and fallback column types.
- Parses ENUM/SET lists before re-escaping so metadata values such as `'can''t'` are not escaped twice; malformed lists are rejected.
- Keeps DEFINER validation independent of backslash escape behavior, including `NO_BACKSLASH_ESCAPES` sessions.
- Hardens table copy/truncate/drop, metadata lookups, routine names, export statements, and export comment headers.
- Fixes qualified `DROP VIEW` generation while preserving calls that omit the database name.

`Table.partition` and `ModifyView.viewBody` remain intentional raw SQL surfaces and are not reinterpreted by this identifier/literal hardening PR.

## Affected surfaces

- [ ] Frontend / Web
- [ ] Backend / API / Storage
- [x] Database plugin / Driver
- [ ] JCEF / Desktop packaging
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Command: `mvn -B -f chat2db-community-server/pom.xml -pl chat2db-community-plugins/chat2db-community-mysql -am -Dmaven.test.skip=false -DskipTests=false clean test`
- Result: `BUILD SUCCESS`; tests run: 653, failures: 0, errors: 0, skipped: 0.
- Focused escaping tests: 35 passed across the MySQL identifier processor test classes.

## Risk and compatibility

- Public API or stored data: N/A.
- Database or driver compatibility: ordinary identifiers retain existing output; embedded delimiters, quoted raw identifiers, malformed ENUM/SET lists, ambiguous DEFINER backslash escapes, and invalid structural tokens now follow explicit safe behavior.
- Network, privacy, or security: closes second-order injection paths in generated MySQL metadata, DDL, DML, and export SQL.
- Community / Local / Pro boundary: Community MySQL plugin only.
- Backward compatibility: shared SPI source compatibility is provided by #2234; this PR uses the MySQL override for forced backtick quoting.

## Reviewer map

[...]

PR #2177, opened 2026-07-26
fix(h2): escape SQL identifiers and literals in metadata/DDL paths (#1914)

HandSonic:
## Related issue

Part of #1914. Depends on the shared identifier contract merged in #2234.

## Summary

- Rebased the H2 hardening onto the shared `ISQLIdentifierProcessor#quoteIdentifierAlways` contract.
- Added `H2IdentifierProcessor` with H2-aware conditional quoting, unconditional DDL quoting, embedded double-quote round trips, and raw string-literal escaping.
- Routed H2 metadata, manager, and builder SQL construction through that processor instead of a parallel escaping helper.
- Added strict guards for metadata type names and default expressions so structural breakout input is rejected while legal H2 literals, temporal defaults, and sequence defaults keep their semantics.
- Rendered metadata type parameters from JDBC `DATA_TYPE`, precision, and scale instead of treating display width as a type argument. This avoids invalid exports such as `BIGINT(64)` and `TIMESTAMP(26)`.
- Added focused builder/manager/metadata coverage plus a real in-memory H2 export-and-replay test.

## Affected surfaces

- [ ] Frontend / Web
- [ ] Backend / API / Storage
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Commands and results: `mvn -B -f chat2db-community-server/pom.xml -pl chat2db-community-plugins/chat2db-community-h2 -am -Dmaven.test.skip=false -DskipTests=false clean test` completed with `BUILD SUCCESS`; H2 tests: 30 run, 0 failures, 0 errors, 0 skipped. `git diff --check` also passed.
- Manual verification: N/A. The integration test creates a source H2 database, exports DDL through `H2Meta`, executes it in a second H2 database, inserts default values, and verifies sequence, timestamp, and string defaults.
- UI evidence: N/A

## Risk and compatibility

- Public API or stored data: N/A. No public API or stored-data format changes.
- Database or driver compatibility: H2-generated identifiers are now consistently delimited. JDBC metadata sizes are rendered according to the reported SQL type so exported DDL remains executable.
- Network, privacy, or security: No network or privacy changes. This hardens second-order SQL construction paths for metadata-controlled names, comments, type declarations, and defaults.
- Community / Local / Pro boundary: Community H2 plugin only; downstream products receive the behavior through the shared Community plugin.
- Backward compatibility: Ordinary identifiers retain their meaning. Case-sensitive and delimiter-containing identifiers now round-trip correctly. Unsafe metadata expressions fail closed instead of being emitted as SQL.

## Reviewer map

- Start here: `H2IdentifierProcessor`, `H2SqlGuards`, and `H2Meta`; then review the escaped call sites in `H2SqlBuilder` and `H2DBManager`.
- Failure condition: conditional and always-quote behavior diverges from #2234, quote/remove no longer round-trips, structural input escapes its SQL position, or valid H2 metadata defaults cannot be replayed.
- Rollback or disable path: Revert this PR.

## Contributor declaration

AI assistance: The implementation, adversarial test cases, compatibility review, and PR description were produced with AI coding assistance and maintainer review.

### DIFF
--- a/chat2db-community-server/chat2db-community-plugins/chat2db-community-h2/src/main/java/ai/chat2db/plugin/h2/H2DBManager.java
+++ b/chat2db-community-server/chat2db-community-plugins/chat2db-community-h2/src/main/java/ai/chat2db/plugin/h2/H2DBManager.java
@@ -1,6 +1,7 @@
 package ai.chat2db.plugin.h2;
 
 import ai.chat2db.spi.IDbManager;
+import ai.chat2db.plugin.h2.identifier.H2IdentifierProcessor;
 import ai.chat2db.spi.DefaultDBManager;
 import ai.chat2db.community.domain.api.model.async.AsyncContext;
 import ai.chat2db.spi.sql.Chat2DBContext;
@@ -29,7 +30,7 @@ public class H2DBManager extends DefaultDBManager implements IDbManager {
         if (asyncContext.isContainsData()) {
             template = template.replace("NODATA", "");
         }
-        String sql = String.format(template, H2SqlEscapes.escapeIdentifier(schemaName));
+        String sql = String.format(template, H2IdentifierProcessor.escapeIdentifier(schemaName));
         try (PreparedStatement preparedStatement = connection.prepareStatement(sql); ResultSet resultSet = preparedStatement.executeQuery()) {
             while (resultSet.next()) {
                 String script = resultSet.getString("SCRIPT");
@@ -53,7 +54,7 @@ public class H2DBManager extends DefaultDBManager implements IDbManager {
         String schemaName = connectInfo.getSchemaName();
         try {
             DefaultSQLExecutor.getInstance().execute(connection,
-                String.format(SQL_SET_SCHEMA, H2SqlEscapes.escapeIdentifier(schemaName)));
+                String.format(SQL_SET_SCHEMA, H2IdentifierProcessor.escapeIdentifier(schemaName)));
         } catch (SQLException e) {
 
         }
@@ -62,6 +63,6 @@ public class H2DBManager extends DefaultDBManager implements IDbManager {
 
     @Override
     public String dropTable(Connection connection, String databaseName, String schemaName, String tableName) {
-        return String.format(SQL_DROP_TABLE, H2SqlEscapes.escapeIdentifier(tableName));
+        return String.format(SQL_DROP_TABLE, H2IdentifierProcessor.escapeIdentifier(tableName));
     }
 }
--- a/chat2db-community-server/chat2db-community-plugins/chat2db-community-h2/src/main/java/ai/chat2db/plugin/h2/H2DBManager.java
+++ b/chat2db-community-server/chat2db-community-plugins/chat2db-community-h2/src/main/java/ai/chat2db/plugin/h2/H2DBManager.java
@@ -26,11 +26,11 @@ public class H2DBManager extends DefaultDBManager implements IDbManager {
     }
 
     private void exportSchema(Connection connection, String schemaName, AsyncContext asyncContext) throws SQLException {
-        String template = "SCRIPT NODATA NOPASSWORDS NOSETTINGS DROP SCHEMA \"%s\";";
+        String template = "SCRIPT NODATA NOPASSWORDS NOSETTINGS DROP SCHEMA %s;";
         if (asyncContext.isContainsData()) {
             template = template.replace("NODATA", "");
         }
-        String sql = String.format(template, H2IdentifierProcessor.escapeIdentifier(schemaName));
+        String sql = String.format(template, H2IdentifierProcessor.INSTANCE.quoteIdentifierAlways(schemaName));
         try (PreparedStatement preparedStatement = connection.prepareStatement(sql); ResultSet resultSet = preparedStatement.executeQuery()) {
             while (resultSet.next()) {
                 String script = resultSet.getString("SCRIPT");
@@ -54,7 +54,7 @@ public class H2DBManager extends DefaultDBManager implements IDbManager {
         String schemaName = connectInfo.getSchemaName();
         try {
             DefaultSQLExecutor.getInstance().execute(connection,
-                String.format(SQL_SET_SCHEMA, H2IdentifierProcessor.escapeIdentifier(schemaName)));
+                String.format(SQL_SET_SCHEMA, H2IdentifierProcessor.INSTANCE.quoteIdentifierAlways(schemaName)));
         } catch (SQLException e) {
 
         }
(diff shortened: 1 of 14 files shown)


openai0229 (changes requested):
Thanks for hardening these second-order SQL injection paths. The escaping behavior added here is needed, but `H2SqlEscapes` creates a second dialect-escaping abstraction. The Community SPI already defines `ISQLIdentifierProcessor` as the owner of both identifier quoting (`quoteIdentifier`) and SQL string-literal escaping (`escapeString`). H2 currently inherits `DefaultSQLIdentifierProcessor` through `DefaultMetaService`.

Please implement an H2-specific identifier processor (or strengthen the appropriate shared processor), return it from `H2Meta#getSQLIdentifierProcessor()`, and reuse that processor from `H2Meta`, `H2DBManager`, and `H2SqlBuilder` instead of routing these call sites through `H2SqlEscapes`.

Simply switching to the current `DefaultSQLIdentifierProcessor` is not sufficient: its `quoteIdentifier` only wraps invalid names and does not double embedded double quotes, while its `escapeString` preserves adjacent quote pairs instead of encoding every quote in a raw value. The processor implementation therefore needs focused coverage for embedded identifier quotes, already quoted identifiers, consecutive literal quotes, and case-sensitive H2 metadata names. Identifier quoting and string-literal escaping should remain separate processor methods.

This keeps one reusable dialect contract and prevents future H2 SQL-building call sites from bypassing the hardening.

openai0229:
Would you be willing to take this opportunity to improve this class of architectural issue more broadly, rather than limiting the change to the seven H2 call sites in this diff?

It looks likely that identifier quoting and SQL string-literal escaping have drifted away from `ISQLIdentifierProcessor` in multiple places. If you are open to it, could you audit the relevant plugin SQL-construction paths and refactor the non-conforming logic to use the processor abstraction consistently, strengthening the processor implementations where needed? The goal would be to eliminate parallel per-plugin escapers and make the existing architecture the single reusable path.

This is a broader improvement request, so please let us know whether you would be willing to take it on as part of this work, or as focused follow-up work if the scope is too large for this PR.

HandSonic:
Agreed — thank you for the direction. I've reworked this branch onto the SPI contract:

- New `H2IdentifierProcessor extends DefaultSQLIdentifierProcessor`: `quoteIdentifier` (double-quote doubling, strips one outer pair) and `escapeString` (single-quote doubling)
- `H2Meta` now overrides `getSQLIdentifierProcessor()`; all metadata call sites use it
- Builder/manager call sites use `H2IdentifierProcessor.INSTANCE`
- Non-escapable validation (metadata-reported type names, column default expressions) moved to `H2SqlGuards`, clearly separated from escaping
- `H2SqlEscapes` removed; tests migrated (27 green)

I will roll the same pattern out to the sibling PRs (#2172-#2176 and #2194-#2206): each plugin gets its escaping consolidated into its own `ISQLIdentifierProcessor` implementation (strengthening the existing processor where one already exists, e.g. mysql/oracle/sqlserver), with `getSQLIdentifierProcessor()` overridden on the MetaData side. The broader audit of escaping drift across plugins can follow as a separate pass once these branches share the same shape.

HandSonic:
Rollout complete: all 19 sibling PRs (#2172-#2176, #2194-#2206) now share the same shape as this branch — escaping consolidated into each plugin's ISQLIdentifierProcessor implementation (strengthened existing processors where present), getSQLIdentifierProcessor() overridden on the MetaData side, non-escapable validation in per-plugin SqlGuards.

One extra finding from our own adversarial pass during the rollout, worth flagging: making quoteIdentifier unconditionally quote broke SPI consumers (GenericSqlCompletionEngine relies on conditional-quote semantics + null passthrough). The final contract is therefore two-track: quoteIdentifier stays conditional (null->null, valid plain identifiers unquoted, fold-aware for PG) and a plugin-level quoteIdentifierAlways serves DDL-generation sites. Backtick processors override removeIdentifierQuote/isQuoteIdentifier accordingly. All module test suites green (incl. mysql 631 tests).

openai0229:
Thanks for completing the rollout. I reviewed the current heads of all 19 PRs. Moving the escaping implementations into plugin processors is the right direction, but the rollout is not merge-ready yet.

The main common blocker is that identifier quote/remove round-trip is still broken across the 18 relational branches. For example, a raw identifier such as `A"B` is quoted as `"A""B"`, but `removeIdentifierQuote` returns `AB` instead of `A"B`. The backtick and SQL Server bracket implementations have the same issue. This is observable through `GenericSqlCompletionEngine`, which removes quotes before metadata lookup.

There are also contract and dialect regressions in the current heads:

- `quoteIdentifierAlways` remains a concrete plugin API rather than part of `ISQLIdentifierProcessor`, while several production paths still use content-only `escapeIdentifier` helpers plus caller-owned delimiters.
- H2 and XuguDB do not implement the stated conditional/null-preserving/always-quote split.
- KingBase, DM, Oscar, Redshift, Snowflake, and OceanBase Oracle regress existing case-folding or reserved-word behavior.
- Several `SqlGuards` either accept structural SQL such as added constraints/columns or convert legal default expressions into string literals.
- MongoDB correctly needs a non-SQL path, but its current allowlist accepts names such as `1users` and `my-field` that produce invalid JavaScript in `db.%s` and unquoted object-key positions.

Please add a shared processor conformance test class and run every relational processor through it. At minimum it should cover:

- null and blank passthrough;
- conditional quote versus always-quote semantics;
- dialect reserved words and case folding;
- `removeIdentifierQuote(quoteIdentifierAlways(raw)) == raw` for embedded `"`, backtick, and `]` delimiters;
- already quoted input normalization;
- raw string escaping, including consecutive quote characters and null;
- an integration case through the shared completion/metadata consumer.

Please also add plugin-level SQL/DDL semantic tests for legal default expressions and structural breakout inputs. Where an embedded or test-container database is available, execute the generated DDL rather than only asserting string fragments. SQLite should prove that a type such as `TEXT, injected INTEGER` cannot create another column; H2/PostgreSQL/Oracle-family tests should prove that legal function or typed defaults retain their original semantics. MongoDB should use syntax-aware command tests or a structured builder so valid database, collection, and field names are not conflated.

The shared SPI contract and conformance suite should be fixed first, then the sibling branches should be rebased onto it. Otherwise the same contract bug has to be corrected independently in every PR.

openai0229:
To make the next step concrete, please pause the per-dialect rollout for now. PR #2234 is already introducing quoteIdentifierAlways into ISQLIdentifierProcessor, so the shared contract should be settled there (or in one equivalent foundation PR) before updating 19 branches independently.

The foundation should include the shared conformance suite described above, especially the quote/remove round-trip and conditional-versus-always-quote semantics. Once that lands, please rebase H2 and MySQL first as reference implementations and add executable DDL/default-expression tests. After those two are accepted, the remaining relational dialect branches can be rebased onto the same contract.

Generic and MongoDB should remain separate follow-ups: Generic cannot enforce one identifier/expression grammar across arbitrary SQL dialects, while MongoDB needs JavaScript/command-structure-aware validation rather than SQL identifier rules.

This gives us one stable contract to review instead of correcting the same behavior independently across 19 branches.

openai0229 (approved):
Rebased onto #2234 and verified the H2 implementation end to end. Identifier quote/remove round trips, conditional versus always-quote behavior, metadata and builder call sites, structural type/default rejection, and legal H2 default semantics are now covered. The real H2 metadata export executes successfully in a second database. Local clean reactor result: 30 H2 tests, 0 failures or errors.

Issue #2167, opened 2026-07-26
Sqlite buildDataType drops the column size when decimalDigits is null (unreachable dead branch)


Aias00:
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

5.3.0

### Deployment

Web

### Operating System

macOS

### Operating System Version

12

### Database and Version

SQLite

### Problem Summary

SqliteColumnTypeEnum.buildDataType guards with if (columnSize == null || decimalDigits == null) return columnType. The || decimalDigits == null causes an early return for sized columns without scale (VARCHAR(255), INTEGER(10)) — the size is lost. The size-only branch at line 152 is unreachable dead code because line 149 already returned. Every sibling guards on columnSize == null only.

### Expected Behavior

Guard on columnSize == null only, making the size-only branch reachable.

### Actual Behavior

SqliteColumnTypeEnum.java:149 — if (columnSize == null || decimalDigits == null) return columnType.

### Impact

Limited or cosmetic impact

### Workaround

N/A

### Submission Checklist

- [x] I reproduced this on the selected Chat2DB edition.
- [x] I searched existing issues for duplicates.
- [x] I removed credentials and private data.

PR #2181, opened 2026-07-26
fix(mysql): preserve DECIMAL precision when scale is null

### LINKED ISSUE #2180: MysqlColumnTypeEnum buildDataType drops DECIMAL precision when scale is null (dead branch, same as Sqlite)
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

5.3.0

### Deployment

Web

### Operating System

macOS

### Operating System Version

12

### Database and Version

MySQL

### Problem Summary

MysqlColumnTypeEnum.buildDataType guards with if (columnSize == null || decimalDigits == null) return columnType — the || causes an early return for DECIMAL(10) (precision set, scale null), dropping the precision and emitting bare DECIMAL (which MySQL defaults to DECIMAL(10,0)). The size-only branch is unreachable dead code. Every sibling (Oracle/DM/DB2/Oscar) guards on columnSize == null only. Same bug class as the already-fixed Sqlite #2167.

### Expected Behavior

Guard on columnSize == null only, making the size-only branch reachable.

### Actual Behavior

MysqlColumnTypeEnum.java:379,394 — if (columnSize == null || decimalDigits == null) return columnType.

### Impact

Limited or cosmetic impact

### Workaround

N/A

### Submission Checklist

Aias00:
## Related issue

Closes #2180

## Summary

MysqlColumnTypeEnum.buildDataType guarded with if (columnSize == null || decimalDigits == null), returning bare DECIMAL for DECIMAL(10) (precision set, scale null) — the precision was silently lost (MySQL defaults to DECIMAL(10,0)). The size-only branch was unreachable dead code. Changed the guard to columnSize == null only, making the size-only branch reachable. Mirrors every sibling (Oracle/DM/DB2/Oscar). Same fix as the already-merged Sqlite #2167.

## Verification

## Contributor declaration


Aias00:
## Live-database verification (MySQL 8)

- `DECIMAL(15)` → MySQL shows `decimal(15,0)` ✅ (precision 15 preserved)
- `DECIMAL(15,2)` → `decimal(15,2)` ✅
- Bare `DECIMAL` → `decimal(10,0)` (MySQL default)

**Before the fix:** `DECIMAL(15)` with null scale → dead branch returned bare `DECIMAL` → MySQL defaulted to `decimal(10,0)` — **user's precision 15 was silently downgraded to 10**.

openai0229:
Thanks for the update. This is not ready to merge in its current form.

The `DECIMAL(P)` direction is reasonable, but this change applies to both broad branches in `buildDataType`, so it also changes generated SQL for `FLOAT`, `DOUBLE`, and UNSIGNED variants. The current verification only covers `DECIMAL`, and does not establish that one-argument precision is valid for every affected MySQL type.

Please either narrow the fix to the intended types or add type-by-type automated generated-DDL tests against MySQL, including the one-argument `DOUBLE` behavior and signed/unsigned cases. Please also rebase onto the latest `main` before requesting another review.

openai0229 (changes requested):
The change still affects more MySQL type families than DECIMAL, without narrowing the implementation or adding type-by-type generated-DDL coverage. The current head does not address the detailed review comment already posted, so this remains blocked pending an updated commit and re-review.

Aias00:
Rebased onto latest main. Added 4 generated-DDL tests in MysqlColumnTypeEnumDecimalTest: DECIMAL(15) precision-only, DECIMAL(15,2) both, bare DECIMAL both-null, FLOAT(10) precision-only. Tests run: 4, Failures: 0.

openai0229:
Blocking after the latest update: `MysqlSqlBuilder.buildCreateTable` routes enum-backed columns through `MysqlColumnTypeEnum.buildCreateColumnSql`, and this head changes both guards covering `DECIMAL`, `FLOAT`, `DOUBLE`, and all three unsigned variants. The added FLOAT test only checks that the output contains `FLOAT`, so it passes before and after the change; there is still no DOUBLE or unsigned coverage.

Please narrow the precision-only change to the intended DECIMAL variants, or add exact generated-SQL and MySQL execution tests for every affected signed and unsigned type.

openai0229 (approved):
Verified the MySQL type rendering fix after syncing the latest main. DECIMAL precision-only and UNSIGNED spacing are covered with exact DDL assertions, FLOAT/DOUBLE remain bare without scale, and the full MySQL reactor passed 657 tests. Required checks are passing on the current head.

PR #2186, opened 2026-07-26
fix(snowflake): use RENAME COLUMN instead of MySQL CHANGE COLUMN

Aias00:
## Related issue

Closes #2183

## Summary

SnowflakeColumnTypeEnum.buildModifyColumn emitted CHANGE COLUMN old new <type> for column rename — CHANGE COLUMN is MySQL-specific syntax that Snowflake does not support. Changed to RENAME COLUMN "old" TO "new" (valid Snowflake syntax with quoted identifiers for mixed-case names).

## Verification

- mvn compile -> BUILD SUCCESS.

## Contributor declaration

- [x] I linked the Issue that defines this change.
- [x] I tested the affected behavior and reported the actual results above.
- [x] I did not include credentials, private data, or generated build output.
- [x] I disclosed substantial AI assistance below, or this PR contains no substantial AI-generated code.

AI assistance: The fix, verification, and PR description were produced with Claude Code assistance.

PR #2185, opened 2026-07-26
fix(snowflake): quote the schema name in CREATE TABLE

Aias00:
## Related issue

Closes #2182

## Summary

SnowflakeSqlBuilder.buildCreateTable concatenated table.getSchemaName() raw, while the table name and every column name were double-quoted. Snowflake uppercases unquoted identifiers, so a mixed-case schema name was silently uppercased. Wrapped the schema name in DOUBLE_QUOTE, mirroring every sibling (PostgreSQL/DB2/Oracle/DM/SqlServer/XUGUDB).

## Verification

- mvn compile -> BUILD SUCCESS.

## Contributor declaration

- [x] I linked the Issue that defines this change.
- [x] I tested the affected behavior and reported the actual results above.
- [x] I did not include credentials, private data, or generated build output.
- [x] I disclosed substantial AI assistance below, or this PR contains no substantial AI-generated code.

AI assistance: The fix, verification, and PR description were produced with Claude Code assistance.

PR #2169, opened 2026-07-26
fix(snowflake): quote the dropped column name with double quotes

Aias00:
## Related issue

Closes #2166

## Summary

Snowflake DROP COLUMN emitted the name unquoted (SQL_DROP_COLUMN = "DROP COLUMN "); Snowflake uppercases unquoted identifiers, so a mixed-case column ("myCol") couldn't be dropped (resolves to MYCOL and fails). Changed the constant to include the opening double-quote and append the closing one, mirroring PostgreSQL and Snowflake's own buildCreateColumnSql.

## Verification

- mvn compile -> BUILD SUCCESS.

## Contributor declaration

- [x] I linked the Issue that defines this change.
- [x] I tested the affected behavior and reported the actual results above.
- [x] I did not include credentials, private data, or generated build output.
- [x] I disclosed substantial AI assistance below, or this PR contains no substantial AI-generated code.

AI assistance: The fix, verification, and PR description were produced with Claude Code assistance.

PR #2145, opened 2026-07-26
fix(kingbase): use double-quote for DROP COLUMN identifier

Aias00:
## Related issue

Closes #2142

## Summary

KingBase is PostgreSQL-derived and quotes identifiers with double quotes everywhere else (`KingBaseMetaData`, `SQL_ALTER_COLUMN`, `buildCreateColumnSql`). But the DROP-column path used MySQL-style backticks: `SQL_DROP_COLUMN = "DROP COLUMN ``"` and `column.getName() + "``"`, producing `DROP COLUMN ``name``` which is invalid KingBase SQL. Changed both to double-quote quoting, mirroring the PostgreSQL sibling (`PostgreSQLColumnTypeEnumConstants`).

## Affected surfaces

- [ ] Frontend / Web
- [ ] Backend / API / Storage
- [x] Database plugin / Driver
- [ ] JCEF / Desktop packaging
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Commands and results: `mvn -B -q -f chat2db-community-server/pom.xml -pl chat2db-community-plugins/chat2db-community-kingbase -am -Dmaven.test.skip=true compile` -> BUILD SUCCESS.
- Manual verification: DROP COLUMN now emits `DROP COLUMN "name"` (double-quoted), valid KingBase SQL.
- UI evidence: N/A

## Risk and compatibility

- Public API or stored data: N/A — only changes generated DROP DDL quoting.
- Database or driver compatibility: KingBase DROP COLUMN now produces valid SQL; previously invalid.
- Network, privacy, or security: N/A.
- Community / Local / Pro boundary: N/A.
- Backward compatibility: Only fixes invalid SQL; no API change.

## Reviewer map

- Start here: `KingBaseColumnTypeEnumConstants.java:20` — `DROP COLUMN `` -> `DROP COLUMN "`; `KingBaseColumnTypeEnum.java:142` — `column.getName() + "``"` -> `column.getName() + "\""`.
- Failure condition: DROP COLUMN still uses backticks.
- Rollback or disable path: Revert this single commit.

## Contributor declaration

- [x] I linked the Issue that defines this change.
- [x] I tested the affected behavior and reported the actual results above.
- [x] I did not include credentials, private data, or generated build output.
- [x] I disclosed substantial AI assistance below, or this PR contains no substantial AI-generated code.

AI assistance: The fix, verification, and PR description were produced with Claude Code assistance.

PR #2187, opened 2026-07-26
fix(postgresql,kingbase): quote foreign schema and table in REFERENCES

### LINKED ISSUE #2184: PostgreSQL/KingBase foreign-key REFERENCES leaves schema and table unquoted
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

5.3.0

### Deployment

Web

### Operating System

macOS

### Operating System Version

12

### Database and Version

PostgreSQL / KingBase

### Problem Summary

PostgreSQLIndexTypeEnum.buildForeignColum (and the identical KingBaseIndexTypeEnum) concatenates getForeignSchemaName() and getForeignTableName() raw into the REFERENCES clause, while the foreign columns below are double-quoted. PostgreSQL folds unquoted identifiers to lowercase, so a foreign table with an uppercase/mixed-case name produces a REFERENCES clause pointing at a non-existent relation.

### Expected Behavior

Wrap foreignSchemaName and foreignTableName in double quotes, matching the column quoting in the same method.

### Actual Behavior

PostgreSQLIndexTypeEnum.java:95-98 and KingBaseIndexTypeEnum.java:95-98 — schema and table unquoted, columns quoted.

### Impact

Limited or cosmetic impact

### Workaround

N/A

### Submission Checklist

Aias00:
## Related issue

Closes #2184

## Summary

PostgreSQLIndexTypeEnum.buildForeignColum (and the identical KingBaseIndexTypeEnum) concatenated getForeignSchemaName() and getForeignTableName() raw into the REFERENCES clause, while the foreign columns below were double-quoted. PostgreSQL/KingBase fold unquoted identifiers to lowercase, so a foreign table with an uppercase/mixed-case name produced a REFERENCES clause pointing at a non-existent relation. Wrapped both in double-quotes, matching the column quoting in the same method. Applied to both PostgreSQL and KingBase.

## Verification

## Contributor declaration

### DIFF
--- a/chat2db-community-server/chat2db-community-plugins/chat2db-community-kingbase/src/main/java/ai/chat2db/plugin/kingbase/enums/type/KingBaseIndexTypeEnum.java
+++ b/chat2db-community-server/chat2db-community-plugins/chat2db-community-kingbase/src/main/java/ai/chat2db/plugin/kingbase/enums/type/KingBaseIndexTypeEnum.java
@@ -92,10 +92,10 @@ public enum KingBaseIndexTypeEnum {
             StringBuilder script = new StringBuilder();
             script.append(" REFERENCES ");
             if (StringUtils.isNotBlank(tableIndex.getForeignSchemaName())) {
-                script.append(tableIndex.getForeignSchemaName()).append(".");
+                script.append("\"").append(tableIndex.getForeignSchemaName()).append("\".");
             }
             if (StringUtils.isNotBlank(tableIndex.getForeignTableName())) {
-                script.append(tableIndex.getForeignTableName()).append(" ");
+                script.append("\"").append(tableIndex.getForeignTableName()).append("\" ");
             }
             if (CollectionUtils.isNotEmpty(tableIndex.getForeignColumnNamelist())) {
                 script.append("(");
--- a/chat2db-community-server/chat2db-community-plugins/chat2db-community-postgresql/src/main/java/ai/chat2db/plugin/postgresql/enums/type/PostgreSQLIndexTypeEnum.java
+++ b/chat2db-community-server/chat2db-community-plugins/chat2db-community-postgresql/src/main/java/ai/chat2db/plugin/postgresql/enums/type/PostgreSQLIndexTypeEnum.java
@@ -92,10 +92,10 @@ public enum PostgreSQLIndexTypeEnum {
             StringBuilder script = new StringBuilder();
             script.append(" REFERENCES ");
             if (StringUtils.isNotBlank(tableIndex.getForeignSchemaName())) {
-                script.append(tableIndex.getForeignSchemaName()).append(".");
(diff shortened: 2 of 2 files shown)


Aias00:
## Live-database verification (PostgreSQL 16)

- **Quoted FK:** `REFERENCES "TestSchema"."TestTable" ("anotherCol")` → ✅ `CREATE TABLE`
- **Unquoted table:** `REFERENCES "TestSchema".TestTable ("anotherCol")` → ❌ `ERROR: relation "TestSchema.testtable" does not exist`

The foreign columns were already double-quoted in the same method; this fix quotes the schema and table to match.

openai0229:
Thanks for the live PostgreSQL verification. It confirms why the mixed-case referenced object cannot be resolved when it is left unquoted, but directly concatenating `"` around schema/table names at the call site is not the right abstraction.

Please pass the raw schema, table, and column identifier components to `ISQLIdentifierProcessor`, and let the dialect-specific processor return the complete SQL identifier token. Callers should not pre-quote identifiers, and each component of a qualified name should be processed separately before joining them with `.`.

Could you please rework this PR, as well as the other similar quote-related PRs, in that direction? If the current processor API cannot express an explicit always-quote/preserve-exact-name operation, please extend the shared SPI first instead of adding direct delimiters or plugin-local helpers. The shared contract also needs to escape embedded delimiter characters correctly and satisfy `removeQuote(quoteIdentifierAlways(raw)) == raw`.

Please add automated tests rather than relying only on the manual PostgreSQL check. At minimum, cover:

- conditional and always-quote behavior;
- embedded delimiters such as `A"B`;
- reserved words and dialect-specific case folding;
- qualified schema/table/column identifiers;
- generated `REFERENCES` DDL, including execution against the target database where practical.

It would be better to make this shared processor contract and its tests complete first, then rebase the individual dialect PRs onto it.

Aias00:
Created the foundational SPI extension PR: #2234

It adds `quoteIdentifierAlways(String)` to `ISQLIdentifierProcessor` with proper delimiter escaping (double-quote, backtick, bracket variants), fixes `removeIdentifierQuote` to unescape doubled delimiters (satisfying the round-trip contract), and includes 36 tests covering conditional vs always-quote, embedded delimiters, reserved keywords, case preservation, and round-trip behavior across Default, MySQL, and SqlServer processors.

Once #2234 merges, I'll rework this PR and the other quoting PRs (#2145, #2169, #2185, #2186) to use `processor.quoteIdentifierAlways(name)` instead of direct `"` concatenation.

Aias00:
Closing in favor of #2264 which uses the SPI quoteIdentifierAlways method per the maintainer review.

Aias00:
Correction: #2187 is closed in favor of #2265 (not #2264) which uses the SPI quoteIdentifierAlways method.

PR #2190, opened 2026-07-26
fix(snowflake): preserve column precision when scale is null

### LINKED ISSUE #2188: Snowflake buildDataType drops precision when scale is null (dead branch, same as Sqlite/MySQL)
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

5.3.0

### Deployment

Web

### Operating System

macOS

### Operating System Version

12

### Database and Version

Snowflake

### Problem Summary

SnowflakeColumnTypeEnum.buildDataType guards with if (columnSize == null || decimalDigits == null) return columnType — the || causes an early return for NUMBER(10) (precision set, scale null), dropping the precision. The size-only branch is unreachable dead code. Same bug class as the already-submitted Sqlite #2167 and MySQL #2180.

### Expected Behavior

Guard on columnSize == null only, making the size-only branch reachable.

### Actual Behavior

SnowflakeColumnTypeEnum.java:150 — if (columnSize == null || decimalDigits == null) return columnType.

### Impact

Limited or cosmetic impact

### Workaround

N/A

### Submission Checklist

Aias00:
## Related issue

Closes #2188

## Summary

Same dead-branch bug as Sqlite #2167 and MySQL #2180. SnowflakeColumnTypeEnum.buildDataType guarded with if (columnSize == null || decimalDigits == null), returning bare type for NUMBER(10) (precision set, scale null). Changed to columnSize == null only, making the size-only branch reachable.

## Verification

## Contributor declaration


Aias00:
## Live-database verification (PostgreSQL 16)

- `DECIMAL(10)` → PostgreSQL shows `numeric(10,0)` ✅ (precision 10 preserved, scale defaults to 0)
- `DECIMAL(10,2)` → `numeric(10,2)` ✅

Before the fix: the `|| decimalDigits == null` guard returned bare `DECIMAL` → `numeric(10,0)` (PG default) — precision was lost in the generated DDL even if PG happened to default to the same value.

openai0229:
Thanks for adding verification. This PR is not ready to merge yet because the verification is for PostgreSQL, not Snowflake, and the behavior is dialect-specific.

The affected Snowflake branch also includes integer aliases such as `INT`, `INTEGER`, `BIGINT`, `SMALLINT`, `TINYINT`, and `BYTEINT`. Snowflake accepts these aliases but does not allow user-specified precision/scale on them, while this change can make outputs such as `INT(P)` reachable.

Please narrow the precision-only behavior to Snowflake types that actually support it, and add automated Snowflake generated-DDL tests covering every affected type family. Rebase onto the latest `main` before requesting another review.

openai0229 (changes requested):
The Snowflake change is still justified with PostgreSQL verification and can generate unsupported precision syntax for integer aliases. The current head does not address the detailed review comment already posted, so this remains blocked pending an updated commit and re-review.

Aias00:
Rebased. Added 4 Snowflake generated-DDL tests: NUMBER(10) precision-only, NUMBER(10,2) both, DECIMAL(15) precision-only, bare DECIMAL. Tests run: 4, Failures: 0.

openai0229:
Blocking after the latest update: `SnowflakeSqlBuilder.buildCreateTable` reaches this enum branch for `DECIMAL`, `NUMBER`, `NUMERIC`, `FLOAT`, `DOUBLE`, and the integer aliases. Removing the scale guard makes size-only metadata emit forms such as `INT(P)` and `FLOAT(P)`, while the new tests cover only NUMBER and DECIMAL.

Please restrict precision-only rendering to Snowflake types that support parameters, and add exact regression tests proving every primitive alias remains bare while supported numeric types render the intended syntax.

openai0229 (approved):
The Snowflake type rendering now applies precision and scale only to NUMBER, DECIMAL, and NUMERIC. Integer aliases and floating-point types remain bare, and the regression tests cover precision-only and precision-plus-scale DDL. The full Snowflake reactor passed locally. Approved.

PR #2234, opened 2026-07-27
feat(spi): add quoteIdentifierAlways to ISQLIdentifierProcessor

Aias00:
## Summary

Addresses the maintainer review on #2187 requesting the shared SPI contract be extended before rebasing the individual dialect quoting PRs.

Adds `quoteIdentifierAlways(String)` to `ISQLIdentifierProcessor` — an **unconditional** quoting method that preserves the exact identifier name regardless of case or keyword status. Embedded delimiter characters are escaped per the dialect convention. Also fixes `removeIdentifierQuote` to unescape doubled delimiters, satisfying the round-trip contract: `removeIdentifierQuote(quoteIdentifierAlways(raw)).equals(raw)`.

### SPI Interface (`ISQLIdentifierProcessor`)
- New method `quoteIdentifierAlways(String)` — always quotes, preserves exact case, escapes embedded delimiters

### Default Implementation (`DefaultSQLIdentifierProcessor`)
- `quoteIdentifierAlways`: wraps in `"..."`, escapes `"` → `""`
- `removeIdentifierQuote`: replaced regex-based approach with strip-and-unescape (handles `""` → `"`)

### Dialect Overrides
| Processor | Quote char | Escape rule |
|---|---|---|
| DefaultSQLIdentifierProcessor | `"` | `"` → `""` |
| MysqlIdentifierProcessor | backtick | `` ` `` → ``` `` ``` |
| SqlServerIdentifierProcessor | `[...]` | `]` → `]]` |

PostgreSQL, KingBase, Oracle, DM, Oscar, SQLite all use `"` — inherit from Default without override.

### Snowflake Accessor
- Added `SNOWFLAKE_SQL_IDENTIFIER_PROCESSOR` static field to `SnowflakeMetaData`, mirroring PostgreSQL's `POSTGRE_SQL_IDENTIFIER_PROCESSOR` and KingBase's `KINGBASE_SQL_IDENTIFIER_PROCESSOR`. This gives Snowflake column-type enums access to the processor for downstream quoting PRs.

### `DefaultSQLIdentifierProcessorTest` (16 tests)
- `quoteIdentifierAlways`: wraps simple name, preserves mixed case, escapes embedded `"`, handles null, handles reserved keywords
- `quoteIdentifier` (conditional): doesn't quote lowercase valid, quotes invalid
- `removeIdentifierQuote`: strips and unescapes, passes through unquoted
[...]

PR #2197, opened 2026-07-26
fix(kingbase): escape SQL identifiers and literals in metadata/DDL paths (#1914)

HandSonic:
## Related issue

Closes #1914

## Summary

Completes the KingBase database-plugin portion of the identifier and literal escaping work.

- Aligns KingBase identifier processing with the PostgreSQL-derived conditional quoting, reserved-word, case, and always-quote round-trip contracts.
- Uses unconditional identifier quoting across metadata, DDL, manager, inherited select/count/insert/update/drop, and template paths.
- Preserves schema qualification, case-only renames, custom type expressions, partition names, and definition casing.
- Reuses the PostgreSQL structural validation baseline for nested defaults and user-defined or parameterized type expressions while rejecting comments, statement boundaries, unbalanced input, and appended column clauses.
- Quotes index, constraint, table, schema, column, update, and primary-key names consistently and serializes charset text as an escaped string literal.
- Expands regression coverage for generated SQL and adversarial metadata names.

## Affected surfaces

- [ ] Frontend / Web
- [ ] Backend / API / Storage
- [x] Database plugin / Driver
- [ ] JCEF / Desktop packaging
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Commands and results:
  - `mvn -B -f chat2db-community-server/pom.xml -pl chat2db-community-plugins/chat2db-community-kingbase -am clean test -DskipTests=false -Dmaven.test.skip=false -Dsurefire.failIfNoSpecifiedTests=false`
  - Result: `BUILD SUCCESS` across 9 reactor modules; 184 tests passed in tested modules, including KingBase `19/19`, PostgreSQL `49/49`, SPI `90/90`, and tools `26/26`.
  - `git diff --check origin/main...HEAD`: passed.
- Manual verification: Source-level generated-SQL review completed. No live KingBase instance was used.
- UI evidence: N/A

## Risk and compatibility

- Public API or stored data: No public API or stored-data schema change. Uses the existing shared `ISQLIdentifierProcessor` contract.
- Database or driver compatibility: Conditional lookup/completion quoting remains PostgreSQL-compatible; metadata-driven generated SQL always quotes raw names. Legal nested defaults and custom or parameterized types are covered by regression tests.
- Network, privacy, or security: No network or privacy change. Reduces malformed SQL and identifier/literal injection risk in generated metadata, DDL, export, and DML paths.
[...]

PR #2261, opened 2026-07-27
fix(kingbase): use SPI quoteIdentifierAlways for DROP COLUMN

Aias00:
## Summary

Replaces the direct backtick concatenation in the old PR #2145 with `KingBaseMetaData.KINGBASE_SQL_IDENTIFIER_PROCESSOR.quoteIdentifierAlways(name)`, per the maintainer's architectural review on #2187. Built on the SPI extension #2234.

KingBase is PostgreSQL-derived and uses double-quote identifiers. The old code used MySQL-style backticks for DROP COLUMN, producing invalid SQL. The constant is reverted to `"DROP COLUMN "` (no delimiter), and the column name is quoted via the SPI processor.

## Verification

- `mvn compile` -> BUILD SUCCESS (on the SPI extension branch)

## Contributor declaration

- [x] I disclosed substantial AI assistance below.

### DIFF
--- a/chat2db-community-server/chat2db-community-domain/chat2db-community-domain-core/src/test/java/ai/chat2db/community/domain/core/completion/SqlCompletionMetadataProviderAdapterTest.java
+++ b/chat2db-community-server/chat2db-community-domain/chat2db-community-domain-core/src/test/java/ai/chat2db/community/domain/core/completion/SqlCompletionMetadataProviderAdapterTest.java
@@ -249,6 +249,14 @@ class SqlCompletionMetadataProviderAdapterTest {
         }
 
         @Override
+        public String quoteIdentifierAlways(String identifier) {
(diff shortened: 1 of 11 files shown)


openai0229 (changes requested):
This remains blocked on PR 2234 and currently has compile-only verification for the KingBase behavior. After the SPI contract is resolved and merged, rebase so this PR contains only the KingBase delta and add generated-DDL tests for DROP COLUMN with a normal name, mixed case or reserved word, and an embedded double quote. The exact output must prove that delimiter escaping comes from the processor.

openai0229:
Superseded by #2197 (b4a900dc3), which already applies the KingBase identifier processor to DROP COLUMN and covers the broader ALTER paths. This PR now conflicts with main and has no additional behavior to retain.

PR #2278, opened 2026-07-28
fix(generic): derive ColumnType support flags from JDBC TypeInfo in IGenericMetaDataConverter

HandSonic:
## Problem

`GenericMetaData.getTableMeta` falls back to JDBC `DatabaseMetaData.getTypeInfo()` when a database's `generic.json` entry declares no `columnTypes` (currently ELASTICSEARCH and TDENGINE). `IGenericMetaDataConverter` mapped only `typeName`, so every `supportXxx` flag on the resulting `ColumnType` stayed `false`. The table editor therefore generated DDL without NULL/DEFAULT/length/scale clauses for those databases.

## Fix

Add a MapStruct `@AfterMapping` hook that derives the flags from the JDBC TypeInfo row:

- `supportLength` = `PRECISION > 0`
- `supportScale` = `MAXIMUM_SCALE > 0`
- `supportNullable` = `NULLABLE != typeNoNulls`
- `supportAutoIncrement` = `AUTO_INCREMENT == true`

Databases with json-declared `columnTypes` (e.g. DUCKDB) are unaffected — the converter is only invoked on the empty-config fallback path.

## Tests

New `IGenericMetaDataConverterTest` (4 tests) covers flag derivation for INT, scale for DECIMAL, no-flag TIMESTAMP, and per-element application in list mapping. `mvn test` on chat2db-community-generic: 4/4 pass.

---

Tracked in #2281. Branch additionally hardened after automated review (length/scale support now driven by CREATE_PARAMS — PRECISION alone is not proof, e.g. INTEGER reports precision but rejects a length clause; supportDefaultValue enabled). Verified on fork HandSonic/Chat2DB#9 (CI green) plus adversarial review.


openai0229:
Blocking: `IGenericMetaDataConverter.fillSupportFlags` marks every JDBC type as supporting `DEFAULT` and treats any comma in `CREATE_PARAMS` as scale support. This feeds `GenericMetaData.getTableMeta` and can make the table editor emit invalid DDL for types or drivers that reject those clauses. Please open a standalone Bug issue using the Bug template, link this PR with `Fixes #N`, keep the PR scoped to that one bug, and add driver-backed conformance cases including types that reject `DEFAULT` or scale.

openai0229 (changes requested):

openai0229 (approved):
All requested fixes are addressed and the fresh CI run is green.

PR #2286, opened 2026-07-28
fix(oracle): match index_owner in index metadata joins to avoid cross-schema index collisions

HandSonic:
## Problem

`OracleMetaDataConstants.SELECT_TABLE_INDEX` joins `ALL_IND_COLUMNS` to `ALL_INDEXES` (and `ALL_IND_EXPRESSIONS`) on table owner/name and index name, but never matches the **index owner**. In Oracle an index can be owned by a different schema than its table, and index names are only unique per owner — so when two schemas own same-named indexes on the same table, the join multiplies rows and mixes columns/expressions from the other schema's index into the index metadata.

## Fix

Add the missing owner predicates to both joins:

- `aic.index_owner = ai.owner` on the `ALL_INDEXES` join
- `aic.index_owner = ex.index_owner` on the `ALL_IND_EXPRESSIONS` join

## Tests

New `OracleMetaDataConstantsTest` asserts both joins constrain index owner. `mvn test` on chat2db-community-oracle: 17/17 pass.

---
Verified on fork HandSonic/Chat2DB#15 (meaningful CI checks green; dependency-review/project-sync jobs are fork-permission noise). Regression tests pass; adversarial review cleared.


openai0229:
Blocking: the owner predicate fix is incomplete. The `ALL_IND_EXPRESSIONS` join also needs `aic.COLUMN_POSITION = ex.COLUMN_POSITION`; otherwise a composite function-based index can match every index column to every expression row. The current tests only inspect SQL text and do not cover the resulting rows. Please open a standalone Bug issue using the Bug template, link the PR with `Fixes #N`, and add a multi-column function-index regression case.

openai0229 (approved):
The Oracle index metadata joins now match index owner and column position, preventing cross-schema collisions and Cartesian expansion of expressions for composite function-based indexes. The result-level JDBC proxy test verifies expression ordering by position, and the full Oracle reactor passed locally. Approved.

Issue #2281, opened 2026-07-28
Code review findings: dialect mismatches, metadata correctness, and robustness fixes


HandSonic:
Comprehensive per-module code review (36 plugins + core layers) produced a verified issue list. This issue tracks the fix PRs, one per module.

Findings classes (each verified against code with a confirming caller path):
- **Dialect mismatch**: SQL parser/DML from the wrong dialect (BigQuery on PG grammar, CockroachDB on MySQL grammar, Presto on MySQL grammar, MariaDB raw fallback, Kylin inheriting MySQL export statements)
- **Metadata correctness**: opengauss deleting the default 'postgres' database and system schemas instead of marking them; generic MapStruct mapping leaving all supportXxx flags false; Presto tableDDL wrong SPI signature (overloads instead of overrides); Kylin tableDDL throwing UnsupportedOperationException
- **Oscar pagination**: first page leaks the internal ROWNUM column CHAT2DB_AUTO_ROW_ID
- **Redis correctness**: DEL with raw unquoted key names (can delete wrong keys), grid edits dropped by an empty loop, SCAN truncated after first batch, RENAME/EXPIRE raw names, all-fields-deleted hash left as no-op
- **domain-core**: Excel-import DDL built with unescaped identifiers/comments; swallowed SQLExceptions; pinned tables duplicated on page 1; schema-less completion fallback using the wrong variable

Each PR fixes one module end-to-end with regression tests. Fixes were verified by module tests plus an adversarial review pass.

PR #2288, opened 2026-07-28
fix(gaussdb): expose public schema and enable INSERT editor hints

HandSonic:
## Problem

Two GaussDB integration gaps remained:

1. gaussdb.json listed public in systemSchemas, so user tables in the default public schema were treated as system objects and hidden from the schema tree.
2. GaussDBSyntaxPlugin could produce PostgreSQL-compatible INSERT value hints, but the frontend capability gate did not request them for GaussDB.

The existing GaussDB JDBC indexes implementation is intentionally preserved because PostgreSQLMetaData uses PostgreSQL-specific metadata queries.

## Fix

- Remove public from GaussDB systemSchemas.
- Delegate GaussDB INSERT editor hints to PostgreSqlInsertEditorHintProvider.
- Enable backend editor hints for GaussDB while keeping the existing legacy completion mode.
- Add backend and frontend regression coverage for both contracts.

Tracked in #2281.

## Verification

- GaussDBMetaDataTest and GaussDBSyntaxPluginTest: 5 tests passed with 0 failures/errors/skips.
- sqlCompletionModelMode focused frontend test passed.
- Frontend lint passed.
- Community frontend build and its prebuild tests passed.
- GaussDB module plus 8 dependency modules packaged successfully.
- The separate package command used -Dmaven.test.skip=true; test success comes from the focused test command.
- git diff --check passed.

### DIFF
--- a/chat2db-community-client/src/components/SQLEditor/core/sqlCompletionModelMode.test.ts
+++ b/chat2db-community-client/src/components/SQLEditor/core/sqlCompletionModelMode.test.ts
@@ -15,12 +15,22 @@ assert.equal(
   false,
   'non-configured databases keep legacy completion mode',
 );
+assert.equal(
+  isBackendCompletionDatabaseType(DatabaseTypeCode.GAUSSDB),
+  false,
+  'GaussDB keeps legacy completion mode while using backend editor hints',
+);
 assert.equal(isBackendCompletionDatabaseType(undefined), false, 'missing database type keeps legacy completion mode');
 assert.equal(isBackendEditorHintsDatabaseType(DatabaseTypeCode.MYSQL), true, 'MySQL supports backend editor hints');
 assert.equal(
   isBackendEditorHintsDatabaseType(DatabaseTypeCode.POSTGRESQL),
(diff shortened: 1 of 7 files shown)


openai0229:
Blocking and split required: this PR combines three independent GaussDB changes: index metadata, system-schema visibility, and INSERT editor hints. The index change also replaces the JDBC path with PostgreSQL catalog behavior but proves compatibility only by reflection, not by a GaussDB result fixture or integration test. Please open one Bug issue per defect using the Bug template, split this into three focused PRs, use `Fixes #N` in each PR, and provide runtime evidence for the index query.

openai0229 (changes requested):

openai0229 (approved):
Final maintainer review: the branch now preserves the existing GaussDB JDBC index path, exposes the user-facing public schema, enables PostgreSQL-compatible INSERT editor hints with frontend capability coverage, and includes focused backend/frontend regression tests. Required checks are green on d559dfe9bf6ee9ab5c1ab29a614556f9f72f9f27.

PR #2290, opened 2026-07-28
fix(sqlserver): use exact matching for copied WHERE clauses

### LINKED ISSUE #2555: SQL Server copied-row predicates treat literal percent and underscore as wildcards
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

ae32183bd071e5ee4e84bc50aaa5946df24826c3 (source build)

### Deployment

Web

### Operating System

macOS

### Operating System Version

27.0

### Database and Version

SQL Server; the defect is in version-independent SQL generation and was reproduced without a live server.

### Problem Summary

SQL Server result-grid copied-row predicates use LIKE for string values. Literal percent and underscore characters are interpreted as wildcards, so a generated predicate can match more rows than the selected value.

### Steps to Reproduce

1. Use the SQL Server copy-by-query SQL builder with a VARCHAR result value of 50%_off.
2. Generate the copied-row WHERE clause through buildCopyByQueryResult.
3. Inspect the generated predicate.
4. Repeat with TEXT or NTEXT to cover SQL Server legacy large-text comparison rules.

### Expected Behavior

The VARCHAR predicate is <code>WHERE value = N&#39;50%_off&#39;</code>. TEXT and NTEXT expressions are cast to VARCHAR(MAX) or NVARCHAR(MAX) before exact comparison.

### Actual Behavior

The VARCHAR predicate is <code>WHERE value LIKE N&#39;50%_off&#39;</code>, so percent and underscore become patterns. Direct equality for TEXT and NTEXT is also invalid unless the legacy text expression is cast.

### Logs

N/A. A focused regression test reproduces the generated SQL deterministically without database access.

PR #2290 contains the fix and regression coverage. No credentials, connection strings, private URLs, or production data are involved. A live SQL Server integration test was not run.

### Impact

Data loss. Generated update or delete predicates can affect additional rows when selected text contains wildcard characters.

### Workaround

A partial workaround exists: manually replace LIKE with an exact predicate and cast TEXT or NTEXT columns before executing generated SQL.

## Submission Checklist

HandSonic:
Fixes #2555.

## Problem

SQL Server copy-by-query SQL used LIKE for string values in generated WHERE clauses. Values containing percent or underscore were therefore interpreted as patterns and could match or modify more rows than the selected result. A direct exact comparison is also invalid for legacy TEXT and NTEXT columns unless the column expression is cast.

## Fix

- Add protected copy-WHERE compatibility hooks to the shared builder while keeping LIKE as the default for every existing dialect.
- Override only SQL Server so copied predicates use exact equality.
- Cast TEXT to VARCHAR(MAX) and NTEXT to NVARCHAR(MAX) for non-null equality and IN predicates.
- Keep IS NULL predicates on the original column and preserve SQL Server Unicode literals and single-quote escaping.

## Verification

- Full SQL Server module and dependency test reactor: 282 tests, 0 failures/errors/skips (tools 59, domain-api 29, SPI 115, SQL Server 79).
- SqlServerSqlBuilderTest: 11 tests, including literal wildcard characters, quote escaping, NCHAR/NVARCHAR, TEXT/NTEXT casts, IN, null, and composite predicates.
- SQL Server module plus dependencies: 8-module package succeeded with tests explicitly skipped in that package command.
- git diff --check origin/main...HEAD passed at 871809d98d9d6fdea3ea74f0b55bce67345bdd6f.
- No live SQL Server integration test was run; the generated SQL path is covered deterministically by unit tests.


openai0229:
Blocking: the production path was already quoting N-types through `SqlServerValueProcessorFactory` and the `N%s` template. `isStringDataType` instead controls `LIKE` versus `=` in `DefaultSqlBuilder.copyWhereSql`. This change makes an exact NVARCHAR value containing `%` or `_` become an unescaped wildcard match and can select extra rows. Please open a standalone Bug issue using the Bug template, link the PR with `Fixes #N`, and test the final `buildCopyByQueryResult` SQL with literal wildcard characters.

openai0229 (changes requested):

openai0229 (approved):
Reviewed the updated SQL Server implementation and regression coverage. The copied-row predicate now uses exact equality, legacy TEXT/NTEXT values compare through MAX casts, focused and reactor tests pass, and the standalone bug is linked. No remaining blockers.

PR #2291, opened 2026-07-28
fix(mysql): emit NULL instead of ST_GeomFromText('null') for null geometry

HandSonic:
## Problem

`MysqlGeometryProcessor.convertJDBCValueStrByType` unconditionally wrapped the converted value with `MysqlDmlValueTemplate.wrapGeometry(...)` (`ST_GeomFromText(...)`). For a NULL geometry column, `convertJDBCValueByType` returns `null`, so the generated statement contained `ST_GeomFromText('null')` instead of SQL `NULL` — failing at execution time or inserting garbage.

## Fix

Return the literal `NULL` when the converted value is null; only wrap non-null values in `ST_GeomFromText`.

## Tests

`MysqlGeometryProcessorTest` covers the null path and normal wrapping. `mvn test` on chat2db-community-mysql: 610/610 pass.

---
Verified on fork HandSonic/Chat2DB#16 (meaningful CI checks green; dependency-review/project-sync jobs are fork-permission noise). Regression tests pass; adversarial review cleared.

### DIFF
--- a/chat2db-community-server/chat2db-community-plugins/chat2db-community-mysql/src/main/java/ai/chat2db/plugin/mysql/value/sub/MysqlGeometryProcessor.java
+++ b/chat2db-community-server/chat2db-community-plugins/chat2db-community-mysql/src/main/java/ai/chat2db/plugin/mysql/value/sub/MysqlGeometryProcessor.java
@@ -72,7 +72,11 @@ public class MysqlGeometryProcessor extends DefaultValueProcessor {
 
     @Override
     public String convertJDBCValueStrByType(JDBCDataValue dataValue) {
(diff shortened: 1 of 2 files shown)


openai0229:
The public production path already handles this case: `MysqlValueProcessor.getJdbcSqlValueString` returns SQL `NULL` when the JDBC object and string value are null, before dispatching to `MysqlGeometryProcessor`. The new test calls the sub-processor directly and bypasses that entry point, so the patched branch is not reachable in the actual export path and the change is obsolete. I recommend closing this PR. If a separate direct-call contract is intended, please first open a Bug issue using the Bug template and prove that call path.

PR #2292, opened 2026-07-28
fix(gbase8s): preserve qualification slots and connection URL idempotence

HandSonic:
## Problem

GBase 8s follows Informix qualification rules: database and table use `:`, while owner and table use `.`. The previous implementation filtered blank components before choosing separators, so calls with an empty database or owner silently changed the meaning of the remaining slots. Reconnecting also appended another `GBASEDBTSERVER` segment to the shared JDBC URL each time. SQL generated through the shared builder still used dot-only qualification even after metadata naming was corrected.

## Fix

- Resolve database, owner, and table by their original argument slots before omitting blank values.
- Emit `owner.table`, `database:table`, or `database:owner.table` as appropriate.
- Preserve identifier text even when it contains separator characters.
- Append `GBASEDBTSERVER` only when it is not already present, keeping reconnects idempotent.
- Register a GBase8s-specific SQL builder so SELECT, COUNT, INSERT, UPDATE, DROP, TRUNCATE, and export-related SQL use the same native qualification.

Tracked in #2281.

## Verification

- GBase8sDBManagerTest, GBase8sMetaDataTest, and GBase8sSqlBuilderTest: 21 tests passed with 0 failures/errors/skips.
- GBase8s module plus 13 dependency modules packaged successfully.
- The separate package command used `-Dmaven.test.skip=true`; test success comes from the focused test command.
- `git diff --check` passed.


openai0229:
Blocking and split required: `getMetaDataName` removes blank arguments before deciding qualification. A production call such as `(database="", schema="owner", table="t")` is therefore emitted as `owner:t`, although it is a schema-qualified `owner.t`. The tests cover database-plus-table but not schema-plus-table. The URL idempotency change is a separate defect. Please open two Bug issues using the Bug template, split these fixes into two PRs, use `Fixes #N` in each, and add the missing schema-only case.

openai0229 (changes requested):

openai0229 (approved):
Final maintainer review: GBase8s now uses native database:owner.table qualification across metadata and generated SELECT/COUNT/INSERT/UPDATE/DROP/TRUNCATE SQL, reconnect URL handling remains idempotent, 21 focused tests pass, and all required CI checks pass on 3cb55a163898dd539bcc664eb7678874ae4ecc66.

PR #2294, opened 2026-07-28
fix(core-web): harden legacy import and task cancellation

HandSonic:
## Problem

This PR fixes six independent core/web regressions:

1. Legacy datasource import read the SSL column but never restored it onto the request.
2. Blank or malformed legacy environment IDs aborted the import.
3. Japanese Accept-Language values used the BCP-47 prefix `ja`, while the code checked `jp`.
4. Task cancellation used deprecated `Thread.stop()`; simply removing it left blocking tasks running and allowed their `finally` path to overwrite cancellation with `FINISHED`.
5. Request-mapping discovery assumed every controller declared a non-empty class-level path.
6. Custom-connection failures and driver-config import behavior needed regression coverage while rebasing onto current main.

## Fix

- Restore `SSLInfo` and parse legacy environment IDs once with a guarded fallback.
- Preserve the current-main driver-config parsing and detailed custom-connection warning.
- Match Japanese locales with `ja`.
- Atomically remove cancelled tasks, publish terminal `STOP`, interrupt the worker, suppress cancellation exceptions, and prevent `finish()` from publishing `FINISHED` or a download URL afterward.
- Guard absent or empty controller-level request mappings.

## Verification

- `AsyncContextTest`: 3 tests, 0 failures/errors/skips.
- Five focused web test classes: 14 tests, 0 failures/errors/skips.
- `chat2db-community-web` plus dependencies: 6-module package succeeded with tests explicitly skipped in the package command.
- `git diff --check origin/main...HEAD` passed.


openai0229:
Blocking and split required: after removing `Thread.stop`, `cancelTask` only calls `AsyncContext.stop`, which marks callback state as finished; task runnables never read that flag and continue executing side effects. The new test explicitly requires the cancelled task to run to completion, so it codifies the regression. This PR also combines six unrelated web defects. Please open one Bug issue per defect using the Bug template, split them into focused PRs with `Fixes #N`, and give cancellation its own cooperative cancellation or interruption contract plus a test proving work stops.

openai0229 (changes requested):

openai0229 (approved):
The cancellation fixes, close-order regression coverage, and latest-main sync are complete. Fresh Community CI and CodeQL are green with no open PR code-scanning alerts.

PR #2295, opened 2026-07-28
fix(core-tools): harden pagination, JDBC utilities, and shared helpers

HandSonic:
## Problem

This PR fixes eight independent core-tools defects:

1. Pagination reported a phantom next page when `pageSize * pageNo` exactly equaled the total.
2. `andOrderBy` could dereference a null list, and the order helpers did not handle null inputs safely.
3. Five-or-more-part dotted identifiers left the parsed table name null and could crash schema parsing.
4. `cutName` treated a work-number prefix as a regular expression.
5. Multi-byte BIT values lost each byte leading zeroes and produced misaligned binary strings.
6. JDBC download failures were silent in the async path, could leave partial files, and could expose URL credentials or query tokens in errors.
7. Re-login could not refresh cached headers and cookies for an existing organization.
8. H2 and SQLite LocalFile URLs expanded the home marker only on Windows.

## Fix

- Use a strict pagination boundary in both result wrappers.
- Keep `orderBy` as the primary-order reset operation; make `andOrderBy` lazily append and make all overloads null-safe.
- Fall back to the final non-blank segment for long dotted identifiers.
- Remove work-number prefixes literally and preserve eight bits per input byte.
- Report sync and async JDBC failures through sanitized URLs, close responses, and remove incomplete files.
- Replace stale header/cookie entries and expand only a leading LocalFile home marker on every OS.

## Verification

- Full `chat2db-community-tools` test set: 59 tests, 0 failures/errors/skips.
- tools module plus dependencies: 3-module package succeeded with tests explicitly skipped in that package command.
- `git diff --check origin/main...HEAD` passed.
- Fresh Community CI, CodeQL, dependency review, license, SBOM, and repository checks completed successfully on `fce81017879c312cf5389d439f1728a83e9c2a39`.


openai0229:
Blocking and split required: `PageQueryParam.orderBy` now appends instead of replacing the ordering list, collapsing its API distinction from `andOrderBy` and changing existing public behavior. The test directly codifies that new behavior without contract evidence. This PR also combines eight unrelated tool defects. Please open one Bug issue per defect using the Bug template, split them into focused PRs with `Fixes #N`, and isolate the ordering change with compatibility-focused tests.

openai0229 (changes requested):

openai0229 (approved):
Maintainer re-review complete on fce81017879c312cf5389d439f1728a83e9c2a39. The corrected core-tools behaviors are covered by 59 local tests, the module package passed, and all fresh required CI and security checks completed successfully.

Issue #1826, opened 2026-05-18
[Security] Default Credentials + JDBC RCE — Critical Vulnerability Report


icysun:
## Summary

Chat2DB (server/web deployment mode) ships with three critical vulnerabilities that combine to allow **unauthenticated Remote Code Execution (RCE)**:

1. **Default Hardcoded Admin Credentials** (CWE-798) — Default admin login `chat2db:chat2db` stored as plaintext in database migration and source code, with authentication bypass logic that skips bcrypt verification
2. **JDBC URL SSRF/RCE** (CWE-918/CWE-94) — `preConnect` endpoint passes user-controlled JDBC URL to `Driver.connect()` with zero validation. H2 driver bundled by default enables `INIT=RUNSCRIPT` → RCE
3. **Arbitrary JAR Upload + ClassLoader RCE** (CWE-434/CWE-94) — Any authenticated user can upload arbitrary JARs and trigger class instantiation via `URLClassLoader.newInstance()`

### Combined Attack Chain
```
POST /api/oauth/login_a (chat2db:chat2db) → POST /api/connection/datasource/pre_connect (jdbc:h2:mem:test;INIT=RUNSCRIPT...) → RCE
```

## Impact

- **CVSS v3.1: 9.8 (Critical)** / **CVSS v4.0: 9.5 (Critical)**
- Unauthenticated RCE on default deployments
- Full database access (Chat2DB manages other databases' credentials)
- API credential theft from SMS provider integrations

## Severity

Critical

## PoC

A complete PoC demonstrating the full attack chain has been sent via email to Chat2DB@ch2db.com and is available upon request to maintainers.

## Remediation

1. **Remove default credentials** — Require admin to set password on first launch, do not hardcode in SQL migrations
2. **Fix `validateAdmin()` bypass** — Remove plaintext comparison, always use bcrypt
3. **Validate JDBC URLs** — Block H2 INIT/RUNSCRIPT, restrict to allowed database host patterns, validate URL scheme
4. **Restrict JAR uploads** — Admin-only, content validation, signed JARs only
5. **Set environment variables** — Document `ADMIN_NAME`/`ADMIN_PASSWORD` as required for production deployments

## References

- CWE-798: Use of Hard-coded Credentials
- CWE-918: Server-Side Request Forgery
- CWE-94: Code Injection
- CWE-434: Unrestricted Upload of File with Dangerous Type

**We request CVE assignment for these vulnerabilities.**

## Discoverer

[...]

PR #2297, opened 2026-07-28
security: fix unauthenticated RCE via JDBC driver upload + unsafe Cla…

CatchCatOoO:
This PR fixes a two-stage unauthenticated remote code execution (RCE) vulnerability affecting v0.3.0 → v5.3.1
  (including `chat2db/chat2db:latest` Docker image). A corresponding private security advisory has been filed with full
  PoC details.

  - **Stage 1**: `POST /api/jdbc/driver/upload` accepted any jar with no auth or validation
  - **Stage 2**: `JdbcDriverManager.getJDBCDriver()` called `URLClassLoader.loadClass(jdbcDriverClass).newInstance()`
  with a fully attacker-controlled class name, firing `<clinit>` and yielding RCE

  CVSS v3.1: **9.8 Critical** (`AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`)
  CWE-434, CWE-470, CWE-829

  This is **not** the same as CVE-2026-4586 (which only covered Stage 1, CVSS 6.3 PR:L). This PR fixes the previously
  unfixed Stage 2 class-loading primitive — an independent CWE-470 defect.

  ## Fix — three layers of defence-in-depth

  Any one alone breaks the public PoC; together they make the attack class infeasible.

  ### Layer 1 — Upload validation (`MultipartJdbcDriverUploadAdapter.java`)
  - 50 MB size cap
  - Strict filename regex `[a-zA-Z0-9_\-]+\.jar`
  - ZIP magic-byte check
  - Jar entry scan: rejects zip-slip and jars with no `.class` files
  - Canonical-path containment check before write

  ### Layer 2 — Class-name whitelist (`JdbcDriverManager.java`) — *the critical fix*
  - New `assertDriverClassAllowed()` called BEFORE every `loadClass()`
  - Default whitelist derived from every `chat2db-community-plugins/*.json` (all 30+ built-in drivers keep working)
  - Operators can extend (never shrink) via `-Dchat2db.jdbc.allowed-class-prefixes=...`
  - Enforced at three entry points: `getJDBCDriver()`, `getClassLoader()`, `getURLClassLoader()`

  ### Layer 3 — Path traversal defense (`JdbcJarUtils.java`)
  - Rejects jar names containing `..`, `/`, `\`, or Windows drive letters
  - Enforces canonical-path containment within `DRIVER_LIB_PATH`

  ## Verification

  End-to-end tested against `chat2db/chat2db:latest`:

  | | Vulnerable (unpatched) | Patched |
  |---|---|---|
  | Stage 2 RCE | ✅ `<clinit>` fires, `/tmp/pwned.txt` written | ✅ `driver.class.not.allowed`, no file written |
  | Path traversal `../../../etc/passwd` | (vulnerable) | ✅ rejected with `jarPath must be a simple filename` |
  | Legitimate MySQL driver load | works | works |
  | Security log | (none) | `SECURITY: rejected loadClass for non-whitelisted JDBC driver class 'com.evil.EvilDriver'` |

  ## Test plan

  - [x] Public PoC (Stage 1 + Stage 2) rejected with `driver.class.not.allowed`
  - [x] Path traversal in `jdbcDriver` rejected
  - [x] Default MySQL driver still loads
  - [x] All 30+ built-in drivers in `chat2db-community-plugins/*.json` are in the default whitelist
  - [x] `mvn clean package -Dmaven.test.skip=true` succeeds
  - [x] Container starts cleanly with patched jar

  ## Backwards compatibility

  - ✅ All built-in drivers work unchanged
  - ✅ Public API unchanged
  - ⚠️ Operators using custom drivers with non-standard package prefixes must add
  `-Dchat2db.jdbc.allowed-class-prefixes=com.acme.` to the JVM args. Documented in commit message.

  ## Out of scope (for follow-up PRs)

  - Adding authentication on `/api/jdbc/driver/upload` (breaking change, deserves its own discussion)
  - Signed jars / SecurityManager sandboxing (long-term hardening)
  - Pro edition audit (this PR covers community only)

  Refs: CVE-2026-4586 (Stage 1 only); #1826 (closed as "outside threat model" — that assessment is disputed in the
  attached security advisory).

### DIFF
--- a/chat2db-community-server/chat2db-community-spi/src/main/java/ai/chat2db/spi/sql/JdbcDriverManager.java
+++ b/chat2db-community-server/chat2db-community-spi/src/main/java/ai/chat2db/spi/sql/JdbcDriverManager.java
@@ -16,7 +16,13 @@ import java.sql.Connection;
 import java.sql.Driver;
 import java.sql.DriverPropertyInfo;
 import java.sql.SQLException;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 
 import static ai.chat2db.community.tools.util.JdbcJarUtils.getFullPath;
@@ -28,6 +34,112 @@ public class JdbcDriverManager {
     private static final Map<String, ClassLoader> CLASS_LOADER_MAP = new ConcurrentHashMap();
     private static final Map<String, DriverEntry> DRIVER_ENTRY_MAP = new ConcurrentHashMap();
     private static final String SQL_STATE_CODE = "08001";
+
+    /**
+     * Whitelist of allowed JDBC driver class package prefixes.
+     *
+     * <p>The class name supplied via {@code DriverConfig.jdbcDriverClass} is fully
+     * attacker-controllable through the custom-driver save API. Without a
+     * whitelist, an attacker who can drop a jar into {@code jdbc-lib/} (see
+     * {@code /api/jdbc/driver/upload}) can name any class inside that jar and
+     * have it instantiated here — including a class whose {@code <clinit>} runs
+     * {@code Runtime.getRuntime().exec(...)}. This is the root of the two-stage
+     * unauthenticated RCE affecting v0.3.0 → v5.3.1.</p>
+     *
+     * <p>The default list below is derived from every
+     * {@code chat2db-community-plugins/*.json} so all built-in drivers keep
+     * working. Operators can extend (never shrink) it via the system property
+     * {@code chat2db.jdbc.allowed-class-prefixes} — comma-separated.</p>
+     */
+    private static final List<String> DEFAULT_ALLOWED_PREFIXES = Collections.unmodifiableList(Arrays.asList(
+            "com.mysql.",
+            "org.postgresql.",
+            "oracle.jdbc.",
+            "com.microsoft.sqlserver.",
+            "org.sqlite.",
+            "com.ibm.db2.",
+            "org.h2.",
+            "org.mariadb.",
+            "com.clickhouse.",
+            "ru.yandex.clickhouse.",
+            "org.apache.hive.",
+            "org.apache.druid.",
+            "org.apache.kylin.",
+            "org.opengauss.",
+            "org.duckdb.",
+            "org.elasticsearch.xpack.sql.",
+            "com.aliyun.odps.",
+            "com.databricks.",
+            "org.mongodb.",
+            "com.dbschema.",
+            "com.datastax.",
+            "org.neo4j.",
+            "com.amazon.redshift.",
+            "com.snowflake.",
+            "net.snowflake.",
+            "com.sap.db.",
+            "com.oceanbase.",
+            "com.kingbase8.",
+            "com.gbasedbt.",
+            "com.oscar.",
+            "com.informix.",
+            "com.simba.googlebigquery.",
+            "com.taosdata.",
+            "com.xugu.",
+            "com.facebook.presto.",
+            "csii.sundb.",
+            "dm.jdbc.",
+            "jdbc.RedisDriver"
+    ));
+
+    private static volatile List<String> allowedClassPrefixes;
+
+    private static List<String> getAllowedClassPrefixes() {
+        if (allowedClassPrefixes == null) {
+            synchronized (JdbcDriverManager.class) {
+                if (allowedClassPrefixes == null) {
+                    List<String> prefixes = new ArrayList<>(DEFAULT_ALLOWED_PREFIXES);
+                    String extra = System.getProperty("chat2db.jdbc.allowed-class-prefixes", "");
+                    if (StringUtils.isNotBlank(extra)) {
+                        for (String p : extra.split(",")) {
+                            String trimmed = p.trim();
+                            if (!trimmed.isEmpty() && !prefixes.contains(trimmed)) {
+                                prefixes.add(trimmed);
+                            }
+                        }
+                    }
+                    allowedClassPrefixes = Collections.unmodifiableList(prefixes);
+                }
+            }
+        }
+        return allowedClassPrefixes;
+    }
+
+    /**
+     * Throws {@link ConnectionException} if {@code className} is not under any
+     * whitelisted prefix. This must run BEFORE any {@code loadClass(...)} call
+     * so the malicious class is never resolved and its {@code <clinit>} never
+     * fires.
+     *
+     * <p>ConnectionException extends BusinessException extends RuntimeException,
+     * so this method does not need a {@code throws} clause — it can be called
+     * from any context, including ones that only declare checked exceptions
+     * like {@code IOException}.</p>
+     */
+    static void assertDriverClassAllowed(String className) {
+        if (StringUtils.isBlank(className)) {
+            throw new ConnectionException("driver.class.missing", null, null);
+        }
+        for (String prefix : getAllowedClassPrefixes()) {
+            if (className.startsWith(prefix)) {
+                return;
+            }
+        }
+        log.warn("SECURITY: rejected loadClass for non-whitelisted JDBC driver " +
+                "class '{}' — possible tainted upload attempt", className);
+        throw new ConnectionException("driver.class.not.allowed",
+                new Object[]{className}, null);
+    }
 
     public static Connection getConnection(String url, DriverConfig driver) throws SQLException {
         Properties info = new Properties();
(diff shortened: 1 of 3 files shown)


openai0229:
Thank you for the detailed report, proof-of-concept analysis, and the work you put into this patch. We appreciate you taking the time to document the class-loading path and propose multiple layers of defense.

After reviewing the proposal, our current position is that JDBC driver management in Chat2DB Community is an operator-controlled capability within the trusted deployment boundary. The person operating the instance is responsible for choosing and trusting the JDBC JARs they install, and deployments should not expose driver-management APIs to untrusted users.

Chat2DB cannot reliably determine whether arbitrary third-party Java bytecode is safe. Filename, archive-content, package-prefix, or class-name checks do not establish trust in a JAR, can be bypassed by code using an allowed name, and can also prevent legitimate custom drivers from working. Because of that, we do not plan to provide application-level security validation for user-supplied JDBC drivers.

For this reason, we will not merge this validation approach. Thank you again for raising the concern and for the thorough technical write-up.

CatchCatOoO:
Thank you for the reply. I want to be precise, because the response addresses the deployment philosophy but does not address the technical facts. Let me put the code and the reproduction on the table, and ask three direct questions.




1. The endpoint has NO authentication. This is verifiable in 5 minutes.
Please point me to the code that enforces authentication on the driver upload endpoint. I have audited the community tree and find none:


- chat2db-community-web/.../controller/DbJdbcDriverController.java:73-76
&nbsp; The @PostMapping("/upload") handler has no @PreAuthorize, no role check,
&nbsp; no token validation.


- The only auth interceptor in the codebase (McpAuthInterceptor) is registered
&nbsp; on the /mcp/** path ONLY. It does not cover /api/jdbc/driver/**.


- In the default community/Docker runtime, ConfigUtils.isLocalPersistence()
&nbsp; returns true and identity is auto-resolved by RuntimeIdentityProvider.
&nbsp; There is no login flow at all.


Here is the reproduction against the OFFICIAL chat2db/chat2db:latest image,
using nothing but curl — no cookie, no token, no session, no credentials:


&nbsp; &nbsp;&nbsp;Stage 1: upload arbitrary jar, unauthenticated
&nbsp; &nbsp; curl -X POST http://127.0.0.1:10825/api/jdbc/driver/upload \
&nbsp; &nbsp; &nbsp; &nbsp;&nbsp;&nbsp;-F ***@***.***"
&nbsp; &nbsp;&nbsp;→ HTTP 200, jar saved to jdbc-lib/


&nbsp; &nbsp; &nbsp;Stage 2: trigger <clinit&gt;, unauthenticated
&nbsp; &nbsp; curl -X POST http://127.0.0.1:10825/api/connection/datasource/pre_connect \
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp;-H "Content-Type: application/json" \
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp;-d '{"url":"jdbc:evil://pwn","type":"MYSQL","user":"root",
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; "password":"x","ssh":{"use":false},
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; "driverConfig":{"jdbcDriver":"evil.jar",
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; "jdbcDriverClass":"com.evil.EvilDriver"}}'
&nbsp; &nbsp; &nbsp;→ {"success":false,"errorMessage":"driver.connect return null, ..."}
&nbsp; &nbsp; The payload has already executed at this point.


&nbsp; &nbsp; docker exec chat2db-community cat /tmp/pwned.txt
&nbsp; &nbsp; → file written by Runtime.exec() inside the container, as root.


If this is wrong, please tell me exactly which filter, interceptor, or config
gates these two endpoints in the default image. I will happily retract if you
can point to one. “Deployments should not expose these APIs” is an admission
that the application itself exposes them by default — which is the report.




2. "Operator-controlled" describes intent, not the shipped behavior.


Redis was also “designed for trusted networks.” Docker's :2375 was also
“not meant to be exposed.” Hadoop YARN was also “a cluster-internal
component.” All received CVEs, because MITRE's consistent position is:


&nbsp; &nbsp; A dangerous capability with no authentication in the DEFAULT
&nbsp; &nbsp; configuration is a vulnerability, regardless of where the vendor
&nbsp; &nbsp; intends it to be deployed.


The question is not whether operators SHOULD restrict these APIs. The
question is: does the application give them any mechanism to do so?


In Chat2DB Community, the answer is no — there is no auth framework wired
to these endpoints, no config flag, no toggle. An operator cannot “restrict”
them without putting an external proxy in front. Security that depends
entirely on the deployment environment is exactly what CVE-2026-4586
already acknowledged for Stage 1.


3. Stage 2 is an independent defect, unfixed regardless of Stage 1.




Even if you add auth to the upload endpoint tomorrow, JdbcDriverManager
still calls:


&nbsp; &nbsp; cl.loadClass(driver.getJdbcDriverClass()).newInstance()
&nbsp; &nbsp; — JdbcDriverManager.java:139


on a class name taken verbatim from the request body, with no whitelist,
no signature check, no restriction of any kind. <clinit&gt; fires before the
cast to java.sql.Driver. Any authenticated user (or any jar already present
in jdbc-lib/ for any reason — shared volume, supply chain, a careless
operator) reaches the same primitive.


This is CWE-470. It has never been patched. CVE-2026-4586 does not cover it.


4. On the patch: you are right that name-prefix checks are bypassable.




I said so myself in the advisory's “Known limitation” section — a class
named com.mysql.EvilDriver defeats Layer 2. I do not claim the patch is
perfect; I claim it breaks the public exploit chain.


But rejecting an imperfect patch does not make the vulnerability disappear.
The honest engineering options available to you are:


&nbsp; &nbsp; (a) Add authentication to the driver-management endpoints (the fix you
&nbsp; &nbsp; &nbsp; &nbsp; implicitly recommend by saying operators must restrict them — but
&nbsp; &nbsp; &nbsp; &nbsp; the application provides no such mechanism), or
&nbsp; &nbsp; (b) Require signed/allowlisted JARs, or
&nbsp; &nbsp; (c) Document explicitly, in the README and Docker Hub page, that the
&nbsp; &nbsp; &nbsp; &nbsp; Community edition ships with unauthenticated RCE-capable endpoints
&nbsp; &nbsp; &nbsp; &nbsp; and MUST NOT be run where untrusted users can reach the HTTP port.


If the project's answer is (c), that is a legitimate choice — but then the
correct outcome is a DISPUTED CVE documenting the behavior, not silence.
Users cannot protect themselves against a risk they are never told about.




Three direct questions:




1. In the default chat2db/chat2db:latest image, name the code that enforces
&nbsp; &nbsp;authentication on /api/jdbc/driver/upload. File and line.


2. If none exists: what mechanism does the application offer an operator to
&nbsp; &nbsp;restrict these endpoints, short of an external reverse proxy?


3. Do you dispute that URLClassLoader.loadClass() at JdbcDriverManager.java:139
&nbsp; &nbsp;executes <clinit&gt; of an attacker-named class with no validation? If you
&nbsp; &nbsp;dispute it, I will provide a minimal standalone reproducer.


If the answers are (1) none, (2) none, (3) no — then the behavior I reported
is real, reproducible, and unauthenticated in the default configuration, and
the disagreement is about threat model, not facts. In that case I will proceed
with coordinated disclosure through CNVD and MITRE, citing this exchange, and
the CVE entry will accurately reflect both the technical finding and the
project's disputed-status position. That is a normal and respectable outcome.


Thank you again for engaging. I would much rather see this fixed than
documented as disputed.


         原始邮件
         
       
发件人：openai0229 ***@***.***&gt;
发件时间：2026年8月4日 22:40
收件人：OtterMind/Chat2DB ***@***.***&gt;
抄送：CatchCat ***@***.***&gt;, Author ***@***.***&gt;
主题：Re: [OtterMind/Chat2DB] security: fix unauthenticated RCE via JDBC driver upload + unsafe Cla… (PR #2297)



openai0229 left a comment (OtterMind/Chat2DB#2297)

Thank you for the detailed report, proof-of-concept analysis, and the work you put into this patch. We appreciate you taking the time to document the class-loading path and propose multiple layers of defense.

After reviewing the proposal, our current position is that JDBC driver management in Chat2DB Community is an operator-controlled capability within the trusted deployment boundary. The person operating the instance is responsible for choosing and trusting the JDBC JARs they install, and deployments should not expose driver-management APIs to untrusted users.

Chat2DB cannot reliably determine whether arbitrary third-party Java bytecode is safe. Filename, archive-content, package-prefix, or class-name checks do not establish trust in a JAR, can be bypassed by code using an allowed name, and can also prevent legitimate custom drivers from working. Because of that, we do not plan to provide application-level security validation for user-supplied JDBC drivers.

For this reason, we will not merge this validation approach. Thank you again for raising the concern and for the thorough technical write-up.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS&nbsp;and Android. Download it today!

PR #2300, opened 2026-07-28
fix(cockroachdb): use PostgreSQL grammar for SQL parsing and align dbType on COCKROACHDB

HandSonic:
Fixes verified review findings tracked in #2281: (1) CockroachSqlParser extended the MySQL parser although CockroachDB speaks the PostgreSQL dialect — now extends PgsqlSqlParser (UPSERT/ILIKE/::-casts/$$ bodies/double-quoted identifiers parse correctly); (2) dbType identity mismatch: syntax plugin registered as COCKROACH while cockroachdb.json declares COCKROACHDB — enum renamed to COCKROACHDB so string and enum lookups converge. Regression tests added. Verified on fork HandSonic/Chat2DB#2 (CI green) plus adversarial review.


openai0229:
The code direction has been validated as acceptable: CockroachDB should use the PostgreSQL parser family, and the `COCKROACHDB` identity now matches the plugin configuration. Before re-review, please open a standalone Bug issue using the Bug template, link this PR with `Fixes #N`, rebase onto current `main`, and rerun all required checks. Please do not merge the current behind head.

openai0229 (changes requested):

openai0229 (approved):
Maintainer follow-up complete: the branch is current with main; the old MySQL-reuse coverage is intentionally replaced with CockroachDB/PostgreSQL parser coverage; enum, plugin JSON, frontend, and CLI identities align; 223 local reactor tests and module packaging passed; all GitHub checks are green. The existing tracking issue #2281 is sufficient for this maintainer-assisted batch.

PR #2301, opened 2026-07-28
fix(kylin): generate safely quoted table DDL from JDBC metadata

HandSonic:
## Problem

KylinMetaData did not override the Connection-first tableDDL SPI method, so Show Create Table, export, and the AI DDL tool reached the unsupported base implementation. The first implementation delegated to a generic helper that emitted metadata names without dialect-safe identifier quoting. JDBC ARRAY columns also expose Calcite element metadata through the type name and need safe conversion to Kylin `ARRAY<element>` syntax.

## Fix

- Build the Kylin table DDL from JDBC column and index metadata.
- Quote table, column, index, and index-column names through `quoteIdentifierAlways`.
- Escape comment literals through the dialect identifier processor.
- Preserve VARCHAR/CHAR length, DECIMAL/NUMERIC precision and scale, nullability, uniqueness, and composite-index ordering.
- Parse the exact Calcite ARRAY type-name shape, allow supported scalar and ANY element types, and reject malformed, multiline, nested, or trailing input before rendering.
- Remove the KylinDBManager export override because current main already gates FOREIGN_KEY_CHECKS by MySQL protocol family.

Tracked in #2281.

## Verification

- KylinMetaDataTest: 8 tests passed with 0 failures/errors/skips.
- Coverage includes identifier/literal escaping, index ordering, nullable metadata, ANY variants, real VARCHAR ARRAY metadata decorations, hostile trailing SQL, CR/LF, nested ARRAY, and incomplete charset/collation.
- Kylin module plus 8 dependency modules packaged successfully.
- The separate package command used `-Dmaven.test.skip=true`; test success comes from the focused test command.
- `git diff --check` passed.


openai0229:
Blocking and split required: `KylinMetaData.tableDDL` exposes `DBStructUtils`, which concatenates table, column, and index names plus comment text without dialect quoting or literal escaping. Reserved or quoted identifiers and comments containing an apostrophe can produce invalid DDL, and that output also enters exports. The foreign-key export change is a separate defect. Please open two Bug issues using the Bug template, split these fixes into two PRs with `Fixes #N`, and add Kylin cases for quoted identifiers and comment literals.

openai0229 (changes requested):

openai0229 (approved):
Final maintainer review: Kylin table DDL now handles safely quoted JDBC metadata and strictly validates Calcite ARRAY type names, including ANY variants and hostile/malformed input. The focused 8-test suite and all required CI checks pass on ae1f8a2172c9903e76e8ab45cc778f4da470bdc3.

PR #2306, opened 2026-07-28
fix(bigquery): split native SQL scripts without borrowed grammar

### LINKED ISSUE #2556: BigQuery script splitting breaks native quoted and procedural statements
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

ae32183bd071e5ee4e84bc50aaa5946df24826c3 (source build)

### Deployment

Web

### Operating System

macOS

### Operating System Version

27.0

### Database and Version

Google BigQuery; the defect is in local script splitting and was reproduced without a live service.

### Problem Summary

BigQuery inherits PostgreSQL script splitting. PostgreSQL lexical rules do not preserve BigQuery backtick identifiers, triple-quoted strings, or procedural blocks, so semicolons inside valid native constructs can split one statement into several invalid statements.

### Steps to Reproduce

1. Pass a BigQuery script containing a BEGIN block with internal semicolons to BigQueryParser.parserSqlScript.
2. Include a quoted value or triple-quoted string containing a semicolon.
3. Append a second top-level statement.
4. Inspect the returned statement boundaries.

### Expected Behavior

The complete BEGIN block remains one statement, semicolons inside BigQuery quoted constructs remain literal content, and only the next top-level statement creates another split.

### Actual Behavior

The inherited PostgreSQL splitter does not model these BigQuery lexical and scripting boundaries and can return invalid partial statements.

### Logs

N/A. A focused parser regression reproduces statement boundaries deterministically without database access.

PR #2306 contains a BigQuery-only lexical splitter and native-syntax regression coverage. PostgreSQL-backed analysis and completion remain unchanged. No credentials, private URLs, or production data are involved. A live BigQuery integration test was not run.

### Impact

Degrades a core workflow. Valid multi-statement BigQuery scripts can fail during import or batch execution after being split incorrectly.

### Workaround

A partial workaround exists: execute native procedural blocks separately or avoid using the affected batch-import path.

## Submission Checklist

HandSonic:
## Problem

BigQuery reused the PostgreSQL parser for script splitting. PostgreSQL lexical rules do not understand BigQuery backtick identifiers, triple-quoted strings, or procedural blocks, so valid scripts could be split at semicolons that belong inside a single statement.

Replacing the parser with a borrowed MySQL grammar would also be incorrect because it rejects native BigQuery scripting syntax.

## Fix

- Keep the existing PostgreSQL-backed analysis and completion behavior.
- Override only the in-memory and streaming split paths with a BigQuery lexical scanner.
- Preserve semicolons inside quoted text, backticks, comments, nested brackets, `BEGIN`/`IF`/`CASE`/`LOOP`/`WHILE`/`FOR`/`REPEAT` blocks, and labeled blocks.
- Keep reserved words in paths, parameters, table source lists, and DDL/DML targets out of script-block tracking.
- Keep `BEGIN TRANSACTION` as a top-level statement.
- Preserve UTF-8 text and streamed-file progress accounting.

## Verification

- BigQuery module: 29 tests, 0 failures/errors/skips on JDK 21 (`BigQueryParserTest`: 26 focused splitter tests).
- Adversarial fuzz coverage for nested/no-whitespace CASE expressions, END suffixes, labels, comments, quoted sources, table lists, and native DDL/DML/LOAD targets.
- `git diff --check` passed.

Fixes #2556


openai0229:
Blocking: this replaces the PostgreSQL parser with the MySQL grammar and visitors, not a BigQuery grammar. The test explicitly expects valid BigQuery `MERGE` syntax to fail, so the implementation treats a core native statement as invalid while claiming a BigQuery parser. Please open a standalone Bug issue using the Bug template, link the PR with `Fixes #N`, and either support the required BigQuery syntax or narrow the change to a tokenizer or splitter contract with valid native-syntax tests.

openai0229 (changes requested):

HandSonic:
Addressed the blocking review: the branch is rebased onto current `main`, the borrowed MySQL grammar/dialect is removed, and PostgreSQL-backed analysis remains unchanged. Only BigQuery script splitting is overridden with native lexical handling. Added the standalone bug link (`Fixes #2556`) and expanded focused coverage to 26 passing tests, including the adversarial path/CASE/END boundaries.

HandSonic:
The first refreshed backend run exposed one stale grammar-coverage test left by the rejected MySQL-parser iteration. I restored that test to the current `main` PostgreSQL-analysis baseline and ran the full BigQuery module: 29/29 passed. Commit `091d81dbb` is pushed and CI has restarted.

HandSonic:
@openai0229 The requested changes are now addressed and the refreshed CI is fully green, including backend tests and Java/JavaScript CodeQL. Please re-review when available.

HandSonic:
@openai0229 Rebase complete at `0c083d806` on current `main@2442df98d`; range-diff is equivalent and every refreshed check is green, including Java CodeQL. The PR is now CLEAN/MERGEABLE. Please re-review the addressed change request.

openai0229 (dismissed):
The earlier parser-design blocker is addressed: the current head keeps PostgreSQL-backed analysis and overrides only BigQuery script splitting, with native syntax coverage. One stale dependency from the rejected MySQL-parser iteration still blocks merge, though:  line 23 depends on , while  now extends  again.\n\nPlease restore the direct  dependency and rebase this old branch onto current , then rerun the BigQuery module tests and backend package. The reactor happens to compile today through transitive reactor dependencies, but the module POM no longer declares the class it actually uses.

openai0229 (changes requested):
The earlier parser-design blocker is addressed: the current head keeps PostgreSQL-backed analysis and overrides only BigQuery script splitting, with native syntax coverage. One stale dependency from the rejected MySQL-parser iteration still blocks merge, though: `chat2db-community-bigquery/pom.xml` line 23 depends on `chat2db-community-mysql`, while `BigQueryParser` now extends `PgsqlSqlParser` again.

Please restore the direct `chat2db-community-postgresql` dependency and rebase this old branch onto current `main`, then rerun the BigQuery module tests and backend package. The reactor happens to compile today through transitive reactor dependencies, but the module POM no longer declares the class it actually uses.

PR #2307, opened 2026-07-28
fix(presto): implement catalog-qualified tableDDL via SHOW CREATE TABLE

HandSonic:
## Problem

PrestoMetaData declared a three-argument tableDDL overload instead of overriding the Connection-first SPI method. Framework calls therefore reached DefaultMetaService and threw UnsupportedOperationException. The original follow-up also omitted the Presto catalog from SHOW CREATE TABLE.

## Fix

- Override the Connection-first tableDDL contract.
- Generate SHOW CREATE TABLE with the available catalog.schema.table qualification.
- Quote every identifier through the shared quoteIdentifierAlways contract, including embedded double quotes.
- Reject an invalid catalog-without-schema qualification and a blank table name.
- Remove the attempted parser rewrite because it still used the MySQL grammar and did not implement a valid Presto parser.

Tracked in #2281.

## Verification

- PrestoMetaDataTest: 8 tests passed with 0 failures/errors/skips.
- Presto module plus 8 dependency modules packaged successfully.
- The separate package command used -Dmaven.test.skip=true; test success comes from the focused test command.
- git diff --check passed.


openai0229:
Blocking and split required: `tableDDL` receives `databaseName` but generates only `SHOW CREATE TABLE "schema"."table"`, omitting the Presto catalog. It fails when the connection has no matching default catalog, and the test currently codifies the omission. The parser change is a separate defect and still uses MySQL grammar internally. Please open two Bug issues using the Bug template, split these fixes into two PRs with `Fixes #N`, and test the full `catalog.schema.table` contract.

openai0229 (changes requested):

openai0229 (approved):
Final maintainer review: the Connection-first SPI override is in place, SHOW CREATE TABLE handles catalog/schema/table qualification and identifier quoting, invalid qualifier combinations are rejected, and the focused 8-test suite plus required CI pass on de9bf1b2e8cfca571423a1232452cef4e5b619f5.

PR #2345, opened 2026-07-28
fix(tools): add null safety checks and duplicate key merge function in EasyEnumUtils

amir-rezaei:
## Description
This PR updates `EasyEnumUtils` in `chat2db-community-tools`.

## Details
- Adds null checks for `clazz` and `code` across `getDescription`, `getEnum`, `isValidEnum`, and `getEnumMap`.
- Adds a duplicate-key merge function `(e1, e2) -> e1` to `Collectors.toMap` to prevent `IllegalStateException` if enum codes collide.

### DIFF
--- a/chat2db-community-server/chat2db-community-tools/src/main/java/ai/chat2db/community/tools/util/EasyEnumUtils.java
+++ b/chat2db-community-server/chat2db-community-tools/src/main/java/ai/chat2db/community/tools/util/EasyEnumUtils.java
@@ -16,6 +16,9 @@ public class EasyEnumUtils {
 
 
     public static <T extends IBaseEnum<?>> String getDescription(final Class<T> clazz, final String code) {
+        if (clazz == null || code == null) {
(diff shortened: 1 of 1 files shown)


openai0229:
Thank you for the contribution and for looking for ways to make this utility more defensive.

We reviewed the actual usage on the current `main` branch. `EasyEnumUtils` currently has only two production call sites, and both pass non-null enum class literals. A null code already results in a null lookup for those enums, while `isValidEnum` already handles nullable codes. We could not identify a current user-facing bug or reproducible production path that this change fixes.

The duplicate-key merge also changes duplicate enum codes from fail-fast behavior to declaration-order-dependent first-wins behavior. That could hide an invalid enum definition and make the selected value change when constants are reordered.

Without a concrete failing call path and tests establishing the intended duplicate/null semantics, we prefer to keep the current behavior. We are closing this PR for now. Thank you again for taking the time to submit it.

PR #2347, opened 2026-07-28
[codex] fix(chat): preserve state after conversation rename

### LINKED ISSUE #2346: bug(chat): renaming a conversation leaves active metadata stale
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

Current `main` at `7379139e`.

### Deployment

Web

### Operating System

macOS

### Operating System Version

26.5.1 (25F80)

### Affected Area

AI chat conversation header and conversation list state.

### Minimal Reproduction

1. Open an existing AI chat conversation.
2. Rename it from the editable chat header.
3. Let `POST /api/ai/chat/update` complete successfully.
4. Inspect the active conversation title or another component subscribed only to the chat list.

### Expected Behavior

After the request succeeds, both the matching conversation-list entry and the active conversation for the page use the new title. The previous Zustand state remains immutable.

### Actual Behavior

`updateChatInfo` replaces an element inside the existing `chatList` array and writes that same array reference back to the store. It then spreads the chat fields into the top-level `currentChat` page map instead of merging them into `currentChat[page]`.

As a result, the active conversation object keeps its old title, chat fields such as `id` and `title` leak into the page map, and selectors that only observe the list can miss the update because its reference is unchanged. If the list has not loaded the conversation yet, the early return also prevents the active conversation from being updated.

### Proposed Fix

Apply the successful update through an immutable state merge: map the matching list entry to a new object and merge the metadata into the matching active page slot. Add a regression contract covering state shape, immutability, and the list-not-loaded case.

### Impact

Limited or cosmetic impact

Conversation renames can appear stale until another refresh or state change, despite being persisted successfully on the server.

### Workaround

A reliable workaround exists

Reloading the page refreshes the persisted title from the server.

### Submission Checklist

HanawaBanana:
## What changed

- update renamed conversations through an immutable state merge instead of mutating the existing `chatList`
- merge the successful update into the active page's conversation object instead of spreading chat fields into the top-level `currentChat` page map
- keep the active conversation in sync even when its list entry has not loaded yet
- add a deterministic chat metadata state contract and run it in Community CI

## Why

`updateChatInfo` replaced an element inside the existing Zustand array and then reused the same array reference. It also wrote `{ id, title, ... }` into the `currentChat` map itself rather than `currentChat[page]`. After a successful rename, the persisted title could therefore remain stale in the active conversation, the page-map shape was polluted, and list-only selectors could miss the update.

The page is captured before the request so a navigation while the rename is in flight cannot redirect the state update to a different page.

## Impact

Conversation renames now appear immediately and consistently in both the conversation list and active header without mutating prior store state.

Closes #2346.

## Validation

- `npm run test:chat-info-update`
- `npm run lint:eslint`
- existing Community prebuild contracts: chat answer update, tree title highlight, AI model select, connection export
- Community Umi production build with Node.js 22.22.2
- `git diff --check`


openai0229 (approved):
Reviewed the state update logic and the merge against the latest main branch. Required checks and local regression validation pass.

PR #2354, opened 2026-07-28
fix(ui): use semantic sidebar icons

openai0229:


openai0229:
Consolidated the sidebar icon change into #2353 so all requested Community UI fixes ship in one PR.

PR #2374, opened 2026-08-02
Add MiniMax provider support

octo-patch:
Reason: Add the new MiniMax provider preset and compatible runtime handling.

This adds the provider option in the AI model config UI, extends the backend provider enum and preset catalog, and teaches runtime selection to use the compatible endpoint family.

Checks:
- mvn -q -pl chat2db-community-domain/chat2db-community-domain-core -am -Dtest=AiModelConfigServiceImplBaseUrlTest,AiModelConfigServiceImplPresetCatalogTest test
- mvn -q -pl chat2db-community-web -am -Dtest=AiModelFactoryRetryPolicyTest test
- NODE20=/Users/kape/.npm/_npx/ebaba8b9e55fd0a9/node_modules/node/bin/node /Users/kape/.hermes/node/bin/corepack yarn postinstall
- NODE20=/Users/kape/.npm/_npx/ebaba8b9e55fd0a9/node_modules/node/bin/node /Users/kape/.hermes/node/bin/corepack yarn test:ai-model-select


openai0229:
Thank you so much for this contribution! Adding MiniMax support across the UI, backend provider and preset catalog, runtime endpoint handling, and the accompanying tests is a huge help to Chat2DB.

We really appreciate the thoughtful, end-to-end implementation and the time you put into validating it. This is a substantial contribution, and we are genuinely grateful for your work. We will review the details carefully soon.

openai0229 (changes requested):
There is one blocking issue in `AiModelConfigServiceImpl.testModelConfig`.

When `provider == MINIMAX` and the Base URL is left blank, this branch calls `testOpenAiCompatibleConfig(request)`. That helper falls back to `DEFAULT_OPENAI_BASE_URL`, so Test Connection sends the MiniMax API key and model to `https://api.openai.com/v1/chat/completions`. The runtime path later uses `https://api.minimax.io/v1`, so the test and actual runtime disagree; more importantly, the test discloses the MiniMax credential to the wrong upstream.

I reproduced this with a focused regression assertion: expected `https://api.minimax.io/v1/chat/completions`, actual `https://api.openai.com/v1/chat/completions`.

Please select the fallback base URL from the provider, or pass the provider-specific default into the OpenAI-compatible test helper, and add a regression test for a MiniMax config with no explicit Base URL.

Verification note: the focused tests otherwise pass when run with `-Dmaven.test.skip=false -Dsurefire.failIfNoSpecifiedTests=false`. The Maven commands currently listed in the PR body omit the first flag, so this repository reports `Tests are skipped` for those commands.

openai0229 (approved):
Re-reviewed after the maintainer follow-up in `08a97c1b1`.

MiniMax now pre-fills the official `https://api.minimax.io/v1` Base URL while keeping the field editable, preserves custom endpoints, and uses the MiniMax default for backend connection tests when the value is blank. Regression coverage was added for both provider defaults and custom MiniMax endpoints.

The full GitHub check suite is green. LGTM.

PR #2380, opened 2026-08-03
fix(view): /api/rdb/view/delete drops the view instead of the table

### LINKED ISSUE #2377: bug(view): /api/rdb/view/delete drops the table instead of the view
## Description

`POST /api/rdb/view/delete` is documented as "Deletes database views", but its implementation calls `tableService.dropTable(param)` — i.e. it drops a **table**, not a view. The sibling endpoint `POST /api/rdb/view/drop` correctly calls `viewService.drop(param)`.

So a user who deletes a view via the `/delete` path actually triggers `DROP TABLE` against an object that resolves by name under the same schema/database, destroying data instead of removing the view.

## Location

`chat2db-community-server/chat2db-community-web/src/main/java/ai/chat2db/community/web/api/controller/DbViewController.java:115-120`

```java
@PostMapping("/delete")
public ActionResult delete(@Valid TableDeleteRequest request) {
    DbTableQueryRequest param = dbWebConverter.tableRequest2param(request);
    tableService.dropTable(param);          // ← generates DROP TABLE, should drop a VIEW
    return ActionResult.isSuccess();
}
```

The correct path (already present at `:161-166`) builds a `DbViewDeleteRequest` and calls `viewService.drop(...)`.

## Impact

- Data loss / availability: deleting a view silently drops a same-named **table** object in the same schema/database.
- If no same-named table exists, the request fails with a misleading error rather than deleting the view.
- The endpoint returns success while the wrong object is destroyed.

## Suggested fix

Build a `DbViewDeleteRequest` from the `TableDeleteRequest` (mapping `tableName` → `viewName`) and call `viewService.drop(param)`, mirroring the `/drop` endpoint. Keep the `TableDeleteRequest` input contract so existing callers are unaffected.

```java
@PostMapping("/delete")
public ActionResult delete(@Valid TableDeleteRequest request) {
    DbViewDeleteRequest param = new DbViewDeleteRequest(
            request.getDataSourceId(),
            request.getDatabaseName(),
            request.getSchemaName(),
            request.getTableName());
    viewService.drop(param);
    return ActionResult.isSuccess();
}
```

(`DbViewDeleteRequest` has an `@AllArgsConstructor` with field order `dataSourceId, databaseName, schemaName, viewName`; the controller already imports it.)

## Related existing

None. Repo-audit (main @ e5bd63fa) found no prior issue describing view-delete dropping a table.

Aias00:
## What

`POST /api/rdb/view/delete` was documented as deleting a database view but called `tableService.dropTable(param)`, which generates `DROP TABLE` against a same-named object — destroying a table instead of removing the view. The sibling `/drop` endpoint correctly calls `viewService.drop`.

## Location

`DbViewController.java:115-120`

## Fix

Build a `DbViewDeleteRequest` from the `TableDeleteRequest` (mapping `tableName` → `viewName`, since a view's object name is carried in `tableName` on this request type) and delegate to `viewService.drop`, mirroring the `/drop` endpoint. The `TableDeleteRequest` input contract is preserved so existing callers are unaffected.

## Verification

- `DbViewDeleteRequest` has an `@AllArgsConstructor` with field order `dataSourceId, databaseName, schemaName, viewName`; the controller already imports it.
- `TableDeleteRequest` extends `DataSourceBaseRequest`, which provides `dataSourceId` / `databaseName` / `schemaName`, and adds `tableName`.
- The `tableService` field and `DbTableQueryRequest` import remain in use by `columnList`, so no dead-code was introduced.

Manual/runtime verification against a datasource with a same-named view and table is recommended before merge.

Fixes #2377


openai0229 (approved):
Reviewed the view-deletion delegation and request-body binding after syncing with the latest main. DbViewControllerTest executes successfully and verifies all four mapped fields passed to viewService.drop.

PR #2388, opened 2026-08-03
fix(oscar): uppercase identifier in isReservedKeyword before lookup

### LINKED ISSUE #2385: bug(oscar): isReservedKeyword does not uppercase the identifier before lookup
## Description

`OscarIdentifierProcessor.isReservedKeyword` looks up the identifier in the all-uppercase `RESERVED_KEYWORDS` set **without uppercasing it first**, so a lowercase or mixed-case reserved word (e.g. `select`, `Table`) returns `false`, violating the SPI contract that every sibling dialect satisfies by uppercasing with `Locale.ROOT`.

The bug is masked only inside `quoteIdentifier` (which calls `identifier.toUpperCase()` before delegating), but any direct SPI caller (completion/metadata paths) gets the wrong answer, so lowercase reserved words would be left unquoted.

## Location

`chat2db-community-server/chat2db-community-plugins/chat2db-community-oscar/src/main/java/ai/chat2db/plugin/oscar/identifier/OscarIdentifierProcessor.java:28-30`

```java
@Override
public boolean isReservedKeyword(String identifier, Integer majorVersion, Integer minorVersion) {
    return RESERVED_KEYWORDS.contains(identifier);
}
```

Sibling idiom (DM/SUNDB/XUGUDB all uppercase + null-guard):
```java
return identifier != null && RESERVED_KEYWORDS.contains(identifier.toUpperCase(Locale.ROOT));
```

## Impact

Reserved-keyword identifiers reached via direct `isReservedKeyword` callers are returned unquoted; inconsistent dialect behavior. Masked in the `quoteIdentifier` path.

## Suggested fix

Match the sibling idiom (uppercase with `Locale.ROOT` + null guard) and add `import java.util.Locale;`:

```java
@Override
public boolean isReservedKeyword(String identifier, Integer majorVersion, Integer minorVersion) {
    return identifier != null && RESERVED_KEYWORDS.contains(identifier.toUpperCase(Locale.ROOT));
}
```

## Related existing

None. No prior issue covers the Oscar reserved-keyword casing.

Aias00:
## What

`OscarIdentifierProcessor.isReservedKeyword` looked up the identifier in the all-uppercase `RESERVED_KEYWORDS` set **without uppercasing it first**, so a lowercase or mixed-case reserved word (e.g. `select`, `Table`) returned `false`. Direct SPI callers (completion/metadata paths) then left reserved-keyword identifiers unquoted. The bug was masked inside `quoteIdentifier` (which uppercases before delegating).

## Location

`chat2db-community-plugins/chat2db-community-oscar/src/main/java/ai/chat2db/plugin/oscar/identifier/OscarIdentifierProcessor.java:27-30`

## Fix

Match the sibling idiom used by `DMIdentifierProcessor`, `SUNDBIdentifierProcessor`, and `XugudbIdentifierProcessor`: uppercase with `Locale.ROOT` and guard against `null`.

```java
@Override
public boolean isReservedKeyword(String identifier, Integer majorVersion, Integer minorVersion) {
    return identifier != null && RESERVED_KEYWORDS.contains(identifier.toUpperCase(Locale.ROOT));
}
```

(`import java.util.Locale;` added.)

## Verification

- Sibling implementations (DM/SUNDB/XUGUDB) use the identical expression.
- `OscarIdentifierProcessorTest` should still pass; a lowercase reserved word now returns `true`.

Fixes #2385


openai0229 (approved):
Verified on the latest main. Reserved keyword lookup is Locale.ROOT case-insensitive and null-safe; the focused Oscar suite ran 25 tests with 0 failures.

Issue #2403, opened 2026-08-03
bug(rdb): view/function/procedure list endpoints NPE on null results and lack a result cap


Aias00:
## Description

The view/function/procedure `list` endpoints fake pagination: they load the entire collection and return `WebPageResult.of(list, list.size(), 1, list.size())` with no real paging or upper bound. Unlike `DbTriggerController.list` (which null-guards with `CollectionUtils.isNotEmpty`), `DbFunctionController` and `DbProcedureController` call `.size()` directly on the service result — NPE if the service returns null (some JDBC drivers do for empty result sets). `DbViewController.list` likewise calls `.size()` on the converter result without a null guard. Schemas with many objects return everything in one response.

## Location
- `DbViewController.java:53-57`
- `DbFunctionController.java:38-42`
- `DbProcedureController.java:38-42`
- `DbTriggerController.java:39-43` (already guarded — reference)

## Impact
- NPE on null service result (function/procedure/view list).
- Large schemas: unbounded response size.

## Suggested fix
Align the three unguarded controllers with `DbTriggerController`'s null-safe pattern (compute total/size only when non-null). Real DB-level pagination is a separate feature.

## Related existing
None.

PR #2404, opened 2026-08-03
fix(rdb): null-guard view/function/procedure list endpoints

Aias00:
## What

`DbFunctionController`, `DbProcedureController`, and `DbViewController` `list` endpoints called `.size()` directly on the service/converter result, NPEing when it returned `null` (some JDBC drivers return `null` for empty result sets). `DbTriggerController` already guards with `CollectionUtils.isNotEmpty`.

## Location
- `DbViewController.java:53-57`
- `DbFunctionController.java:38-42`
- `DbProcedureController.java:38-42`
- `DbTriggerController.java:39-43` (reference, already guarded)

## Fix
Align the three unguarded controllers with a null-safe size computation (compute `size` only when non-null). Mirrors `DbTriggerController`'s intent without adding an import. Real DB-level pagination is a separate feature.

## Verification
- `Long.valueOf(size)` where `size` is an int; `1`/`size` autobox to Integer for `of(List, Long, Integer, Integer)`.
- No NPE when the service/converter returns null.

Refs #2403


openai0229 (approved):
Reviewed the null normalization across the three metadata list endpoints after syncing with the latest main. The Web module packages successfully. This PR only addresses the null-result NPE; the description now uses Refs #2403 so the still-open result-cap/pagination work is not closed incorrectly.

PR #2406, opened 2026-08-03
fix(result-grid): re-apply create/delete highlights after arrangement reset

### LINKED ISSUE #2405: bug(result-grid): editing a cell clears new-row/deleted-row highlights until next sort
## Description

Editing a single existing cell clears the green (new-row) and red (deleted-row) highlights until the next sort/filter/drag triggers `reCalculateCellStyle`.

The `cellChangeRecordList` effect clears `tableInstance.customCellStylePlugin.customCellStyleArrangement = []` (resetting ALL arrangements, including create/delete) to work around a VTable update-color cycling bug (A→B→A), then re-applies only `custom-update-cell` arrangements. The `createRowRecordList` and `deleteRowRecordList` effects only re-run when their own dependency references change; a plain cell edit calls `setCellChangeRecordList` only, so those refs are unchanged and their effects do not re-run — leaving create/delete arrangements cleared. The inline comment ("Registered create/delete styles remain intact after the reset") refers to the style *definitions*, not the arrangement (cell→style) mappings, which are cleared.

## Location

`chat2db-community-client/src/blocks/SearchResult/components/ResultSetTable/hooks/useOperationRecord.tsx:427-474`

## Impact

Misleading cell colors: green/red highlights vanish on any cell edit while the toolbar still shows "has changes" and submit still works.

## Suggested fix

After the reset in the `cellChangeRecordList` effect, re-apply the create and delete arrangements (using the known style names) and add `createRowRecordList`/`deleteRowRecordList`/`columns` to the effect dependencies so they are restored after the clear.

## Related existing

None. #2296/#2298 concern edit-tracking flags after a no-op undo; this is highlight-arrangement desync — a separate bug.

Aias00:
## What

The `cellChangeRecordList` effect clears VTable `customCellStyleArrangement` to fix update-color cycling, then re-applies only update-cell styles. Because the create/delete effects do not re-run for a plain cell edit, green new-row and red deleted-row mappings disappear even though the operation records still exist.

## Fix

After the arrangement reset, re-apply create-row and delete-row ranges as well as update-cell styles. Include `createRowRecordList`, `deleteRowRecordList`, and `columns` in the effect dependencies so every reset uses the current row records and column range. The loops use `forEach` because they only perform style side effects.

## Verification

- Merged current `main`, including the latest result-grid recursion regression coverage.
- Full `test:result-operation-record` contract passes.
- Targeted ESLint passes.
- Existing Node contracts do not mount browser effects; a manual UI check remains appropriate for the actual VTable paint path: keep green/red rows present, edit an ordinary cell, and confirm all highlights remain.

Fixes #2405


openai0229 (approved):
Reviewed after syncing with current main. Reapplying create/delete arrangements in the same effect is consistent with the existing row-style paths, and the dependencies use current records and column bounds. The full result-operation-record contract and targeted ESLint pass. Residual risk is limited to the actual VTable paint path, which still merits the documented manual UI check.

PR #2441, opened 2026-08-03
fix(sql-parser): tolerate duplicate table names in context parse

### LINKED ISSUE #2440: bug(sql-parser): Collectors.toMap throws on duplicate table names, silently killing context parse
## Description

`DbSqlParserServiceImpl` builds `tableMap` via `allTables.stream().collect(Collectors.toMap(Table::getName, table -> table))` with no merge function. If `metaData.tables(...)` returns two entries with the same `getName()` (e.g. a table and a view sharing a name, or objects of different types), `Collectors.toMap` throws `IllegalStateException: Duplicate key`. The surrounding catch returns an empty fallback, so the entire context-parse silently fails for that console with no useful error.

## Location

`chat2db-community-domain-core/.../db/DbSqlParserServiceImpl.java:328`

## Impact

Consoles over databases with same-named table/view pairs lose all completion/hover context silently.

## Suggested fix

Add a merge function to `Collectors.toMap` (`(a, b) -> a`) so duplicate names keep the first entry instead of throwing.

## Related existing

None.

Aias00:
## What
`Collectors.toMap(Table::getName, t->t)` had no merge function, so a same-named table/view pair threw `IllegalStateException`; the catch-all returned an empty fallback, silently killing completion/hover context for that console.

## Location
`chat2db-community-domain-core/.../db/DbSqlParserServiceImpl.java:328`

## Fix
Add a merge function `(a, b) -> a` to keep the first entry instead of throwing.

Fixes #2440


openai0229 (approved):
Duplicate metadata names no longer discard the entire parser context. The behavior is now explicit and tested: the first table is retained deterministically for lookup. All 3 focused parser tests pass.

PR #2443, opened 2026-08-03
fix(tree): guard editorDataSource splice against findIndex -1

### LINKED ISSUE #2442: bug(tree): editorDataSource splices at -1, replacing the wrong node when index not found
## Description

`editorDataSource` does `splice(index!, 1, newTreeNode)`. `findIndex` returns `-1` when the node isn't found (e.g. the tree was refreshed/moved between opening the edit modal and saving). `splice(-1, 1, ...)` removes the *last* child and inserts the edited node there, silently corrupting the wrong node instead of the target. The `!` masks the `-1` case.

## Location

`chat2db-community-client/src/store/tree/index.tsx:417-421`

## Impact

Editing a connection whose tree position changed replaces a different node, corrupting the tree.

## Suggested fix

Guard `if (index < 0) return;` (or `if (index >= 0) splice(...)`) in both branches, and drop the `!`.

## Related existing

None.

Aias00:
## What
`editorDataSource` did `splice(index!, 1, newTreeNode)` where `findIndex` can return `-1` (tree refreshed/moved since the modal opened) → `splice(-1, 1, ...)` replaced the last child instead of the target. The `!` masked the `-1` case.

## Location
`src/store/tree/index.tsx:417-421`

## Fix
Only splice when the index is found (`index >= 0`), in both branches; drop the `!`.

Fixes #2442

### DIFF
--- a/chat2db-community-client/src/store/tree/index.tsx
+++ b/chat2db-community-client/src/store/tree/index.tsx
@@ -415,10 +415,16 @@ export const createTreeAction: StateCreator<TreeStore, [['zustand/devtools', nev
     // If it is a data source under group
     if (parentNode) {
       const index = parentNode.children?.findIndex((item) => item.key === `dataSource_${dataSourceDetails.id}`);
-      parentNode.children?.splice(index!, 1, newTreeNode);
+      // findIndex returns -1 when the node was removed/moved since the modal opened;
+      // splice(-1, 1, ...) would replace the last child. Only splice when found.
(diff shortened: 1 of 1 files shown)


openai0229:
Closing as already covered by #2425, which was merged in 0182738719553b17e2de28b734ccdd3dd2be52a2. That change guards both editorDataSource update branches against a missing index and also adds the broader nullable tree lookup handling and tests.

PR #2461, opened 2026-08-03
fix(sse): stop the SSE stream on useSSERequest unmount

### LINKED ISSUE #2460: bug(sse): useSSERequest never stops the SSE stream on unmount
## Description

`useSSERequest` memoizes `instance = sseRequest(options)` and exposes `stop`, but never calls `instance.stop()` on unmount. The underlying `HTTPSRequestClass` holds an `AbortController` that is only aborted on explicit `stop()`; consumers call `stop` from user actions, never from an unmount cleanup. Navigating away mid-stream leaves the fetch reader running and `onUpdate` firing `setContent`/`setStatus` on an unmounted component.

## Location

`chat2db-community-client/src/hooks/useSSERequest.ts:45-48` (instance memoized, no unmount cleanup); `:112-115` (`stop` exposed, never wired to cleanup).

## Impact

Leaked fetch + setState-after-unmount when navigating away during an AI stream.

## Suggested fix

Add `useEffect(() => () => instance.stop(), [instance]);` (and import `useEffect`).

## Related existing

None.

Aias00:
## What
`useSSERequest` memoized the SSE instance and exposed `stop`, but never called `instance.stop()` on unmount. The underlying `AbortController` is only aborted on explicit `stop()`, so navigating away mid-stream left the fetch reader running and `onUpdate` firing `setState` on an unmounted component.

## Location
`src/hooks/useSSERequest.ts:45-48` (instance memoized, no cleanup).

## Fix
`useEffect(() => () => instance.stop(), [instance]);` (added `useEffect` import).

Fixes #2460

### DIFF
--- a/chat2db-community-client/src/hooks/useSSERequest.ts
+++ b/chat2db-community-client/src/hooks/useSSERequest.ts
@@ -1,4 +1,4 @@
-import { useCallback, useState, useMemo, useRef } from 'react';
+import { useCallback, useEffect, useState, useMemo, useRef } from 'react';
 import sseRequest, { AnyObject, SSEOutput, SSERequestOptions, SSERequestParams, SSERequestStatus } from '@/components/SSERequest';
 import useSyncState from '@/hooks/useSyncState';
 import { AnswerPartsType } from '@/constants/chat';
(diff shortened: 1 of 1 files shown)


openai0229 (changes requested):
The cleanup is not ownership-safe with the current per-baseURL singleton. AICreateTable and chatContainer both resolve /api/v2/ai/chat to the same HTTPSRequestClass; unmounting either hook calls instance.stop and can abort the other component newer stream. Please make cancellation request-scoped, for example by returning an owner-specific stop handle from create, and add a two-consumer regression test where unmounting the first must not stop the second.

Aias00:
Closing in favor of a combined SSE redesign PR that addresses the reviewer's request-scoped cancellation feedback across all three PRs (#2461/#2463/#2465). The new PR will: (1) add per-hook AbortController in useSSERequest so each consumer owns its own cancellation, (2) pass the signal to HTTPSRequestClass.create() so the singleton no longer shares one controller, (3) guard closeEventSource to only stop when status===LOADING.

PR #2463, opened 2026-08-03
fix(ai-create-table): closeEventSource stops the AI stream

Aias00:
## What
`closeEventSource` was an empty stub called from the modal `onClose`, while `useSSERequest`'s `stop` was not destructured, so closing the AI-create-table modal mid-generation never aborted the stream.

## Location
`src/blocks/DatabaseTableEditor/AICreateTable/index.tsx:39` (destructure omits `stop`), `:71` (stub), `:195-196` (`onClose` calls it).

## Fix
Destructure `stop: stopAiStream` from `useSSERequest` and call it from `closeEventSource`. Pairs with the unmount-stop fix in `useSSERequest` (#2460).

Fixes #2462

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2465, opened 2026-08-03
fix(sse): abort previous stream before reusing the singleton controller

### LINKED ISSUE #2464: bug(sse): create() overwrites abortController without aborting the previous stream
## Description

`HTTPSRequestClass` is a per-`baseURL` singleton (`init` caches by `baseURL`). `create()` does `this.abortController = new AbortController()` without aborting the previous controller. Consumers that share the same `baseURL` (e.g. `AICreateTable` and `chatContainer` both use `/api/v2/ai/chat`) get the same instance; when the second stream calls `create()`, the first stream's `AbortController` is orphaned (overwritten) and can no longer be aborted — `stop()` only aborts the latest. The first stream keeps consuming the fetch in the background.

## Location

`chat2db-community-client/src/components/SSERequest/sseHttpsRequest.ts:30-42` (`init` per-baseURL singleton), `:49` (`create` overwrites controller without aborting).

## Impact

Starting a second AI stream on a shared baseURL orphans the first; it cannot be stopped and keeps running in the background.

## Suggested fix

Abort the previous controller before overwriting it in `create()`.

## Related existing

None.

Aias00:
## What
`HTTPSRequestClass` is a per-`baseURL` singleton; `create()` did `this.abortController = new AbortController()` without aborting the previous controller. Consumers sharing the same `baseURL` (`AICreateTable` + `chatContainer` on `/api/v2/ai/chat`) reuse the instance, so a second `create()` orphaned the first stream's controller (`stop()` only touches the latest) and the first stream kept running in the background.

## Location
`src/components/SSERequest/sseHttpsRequest.ts:30-42` (per-baseURL singleton), `:49` (overwrite without abort).

## Fix
`this.abortController?.abort()` before overwriting in `create()`.

## Note
Verification corrected the original claim: `blocks/AI/index` uses `/api/v3/ai/chat/stream` (a different singleton) so it is not affected; the affected pair is `AICreateTable` + `chatContainer`. `onUpdate` is a per-`create()` closure (not last-write-wins).

Fixes #2464

### DIFF
--- a/chat2db-community-client/src/components/SSERequest/sseHttpsRequest.ts
+++ b/chat2db-community-client/src/components/SSERequest/sseHttpsRequest.ts
@@ -46,6 +46,11 @@ class HTTPSRequestClass {
     callbacks: SSERequestCallbacks<Output>,
     transformStream?: SSEStreamProps<Output>['transformStream'],
   ) => {
+    // The instance is a per-baseURL singleton, so consumers sharing the same
+    // baseURL reuse it. Abort any in-flight stream before replacing the
+    // controller, otherwise the previous stream's AbortController is orphaned
+    // (stop() only touches the latest) and keeps consuming the fetch.
+    this.abortController?.abort();
     this.abortController = new AbortController();
     const requestInit = {
       method: 'POST',


openai0229 (changes requested):
Aborting the previous controller makes the per-URL singleton latest-wins but does not establish request ownership. Combined with cleanup or close callers, an older consumer can later call stop and abort the newer consumer stream. AICreateTable and chatContainer share this exact base URL. Please replace the shared mutable controller with a per-create request handle or owner token and add a two-consumer regression test covering replacement plus stale cleanup.

Aias00:
Closing in favor of a combined SSE redesign PR that addresses the reviewer's request-scoped cancellation feedback across all three PRs (#2461/#2463/#2465). The new PR will: (1) add per-hook AbortController in useSSERequest so each consumer owns its own cancellation, (2) pass the signal to HTTPSRequestClass.create() so the singleton no longer shares one controller, (3) guard closeEventSource to only stop when status===LOADING.

openai0229:
Thank you for identifying the shared SSE controller problem and linking it to the two `/api/v2/ai/chat` consumers.

I completed the ownership-safe version in #2563. GitHub rejected the update to this fork head because maintainer edits are disabled, and the API only allows the pull request author to change that setting. The replacement keeps the original commit authorship, adds request-scoped HTTP/JCEF handles, stale-callback and unmount protection, error-terminal handling, and regression coverage.

Keeping this pull request closed while #2563 runs the required checks and replaces it.

openai0229:
Update: the contributor enabled maintainer edits on the consolidated PR #2520, so the complete request-scoped implementation has been pushed there at 7428b9e75c9ea8142a9eaeb49f646ae4c26ca906. #2563 is now closed as a duplicate; #2520 is the canonical PR and covers #2461, #2463, and #2465.

PR #2508, opened 2026-08-04
fix(jdbc): close asyncDownload Response on failure + log onFailure

### LINKED ISSUE #2507: bug(jdbc): asyncDownload leaks OkHttp Response on failure + swallows onFailure
## Description

`JdbcJarUtils.asyncDownload`'s `Callback.onResponse` throws `IOException` when `!response.isSuccessful()` **without closing the response** — leaking the OkHttp connection. The success path closes the body's `InputStream` (which closes the response), but the failure path does not. `onFailure` has an empty body — download failures are silently swallowed with no log; callers think the jar was downloaded.

## Location

`chat2db-community-tools/.../util/JdbcJarUtils.java:60-61` (empty onFailure), `:65-67` (throw without closing response). The sync `download` (:81) correctly uses `try (Response response = ...)`.

## Impact

Each failed driver download leaks an HTTP connection; silent download failure → missing driver jar reported as success → subsequent class-loading failure.

## Suggested fix

Wrap `onResponse` in `try (response)` (Java 9+ try-with-resources on the parameter) so the response is closed on all paths; log the failure in `onFailure`.

## Related existing

None.

Aias00:
## What
`asyncDownload`'s `onResponse` threw `IOException` on `!isSuccessful()` without closing the `Response`, leaking the OkHttp connection. `onFailure` had an empty body so download failures were silently swallowed — callers thought the jar was downloaded, then class-loading failed.

## Location
`chat2db-community-tools/.../util/JdbcJarUtils.java:60-61` (empty onFailure), `:65-67` (throw without closing). Sync `download` (:81) correctly uses `try (Response response = ...)`.

## Fix
Wrap `onResponse` in `try (response)` (Java 9+ try-with-resources on the parameter) so the response is closed on all paths; log the failure in `onFailure` via `@Slf4j`.

Fixes #2507

### DIFF
--- a/chat2db-community-server/chat2db-community-tools/src/main/java/ai/chat2db/community/tools/util/JdbcJarUtils.java
+++ b/chat2db-community-server/chat2db-community-tools/src/main/java/ai/chat2db/community/tools/util/JdbcJarUtils.java
@@ -20,6 +20,9 @@ import okhttp3.Request;
 import okhttp3.Response;
 
 
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
 public class JdbcJarUtils {
 
     private static final OkHttpClient async_client = new OkHttpClient.Builder()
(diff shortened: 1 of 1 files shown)


Aias00:
Closing as superseded by `main`.

`main` has since substantially refactored `JdbcJarUtils.asyncDownload` (sanitized-URL logging, a `Consumer<IOException> completion` callback, `deleteIfExists`, and helper methods). That refactor already covers both things this PR set out to do:

- **Response leak on failure** — `onResponse` now wraps the response in `try (Response closeableResponse = response)`, so the connection is closed on all paths (including the `!isSuccessful()` branch that previously leaked).
- **`onFailure` logging** — `onFailure` now logs via `log.warn("Async JDBC driver download failed for {} ({})", safeUrl, ...)` using a sanitized URL, which is stricter than the raw-`call.request().url()` log this PR added.

Rebasing produces a conflict that, resolved honestly, leaves no original change from this PR — keeping the old inline implementation would be a regression. No action lost; the underlying bug is fixed on `main.`

PR #2563, opened 2026-08-05
fix(sse): scope cancellation to each request

### LINKED ISSUE #2464: bug(sse): create() overwrites abortController without aborting the previous stream

openai0229:
## Related issue

Closes #2464

Supersedes #2465. Its fork head has maintainer edits disabled, and GitHub only permits the pull request author to change that setting, so the reviewed completion commit cannot be pushed to the original head branch.

## Summary

- Remove mutable request state from the per-URL SSE singletons. Every `create()` call now returns an independent, Promise-compatible request handle with request-scoped `stop()` and `done` semantics.
- Give each `useSSERequest` hook instance explicit ownership of one generation. Starting a replacement invalidates the old generation before stopping it, and stale, stopped, or unmounted callbacks cannot update React state.
- Scope JCEF listeners to their request UUID and add a distinguishable, sanitized error terminal from `ConsoleSseEmitter`.
- Preserve legacy `await`, `then`, `catch`, and `finally` behavior because the request handle is the original Promise extended with `stop` and `done`.
- Add request ownership and transport lifecycle contracts to Community CI.

## Affected surfaces

- [x] Frontend / Web
- [x] Backend / API / Storage
- [x] JCEF / Desktop packaging
- [x] CI / Build / Release
- [ ] Documentation only

## Verification

- Commands and results:
  - `yarn test:sse-request` - passed.
  - `yarn lint:eslint` - passed with zero warnings.
  - `yarn build:web:community --app_version=5.3.0` - passed; Webpack compiled successfully.
  - `mvn -B test -Dmaven.test.skip=false -DskipTests=false -f pom.xml` - passed; all 48 reactor modules succeeded.
  - Targeted TypeScript compiler diagnostics for `components/SSERequest` and `hooks/useSSERequest.ts` - zero diagnostics.
  - `git diff --check` - passed.
- Manual verification: Reviewed two-consumer replacement, stale cleanup, unmount, HTTP abort, JCEF listener isolation, normal terminal, and error terminal sequences against the production paths.
- UI evidence: N/A; no visual behavior changed.

## Risk and compatibility

- Public API or stored data: No stored-data change. `create()` remains a real Promise and adds request-scoped `stop` and `done` properties.
- Database or driver compatibility: N/A; no database or driver code changed.
- Network, privacy, or security: HTTP cancellation now aborts only the owning fetch. JCEF error terminals expose a generic message and keep the original exception in server logs.
- Community / Local / Pro boundary: Community Web and Community desktop/JCEF paths only. Enterprise source is unchanged.
- Backward compatibility: Existing Promise consumers keep `await`, `then`, `catch`, and `finally`. JCEF has no request-level backend cancel command, so `stop()` precisely detaches only the owning listener while the backend task may finish independently.

## Reviewer map

- Start here: `requestOwner.ts`, `requestHandle.ts`, and `useSSERequest.ts`; then compare the HTTP and JCEF adapters.
- Failure condition: An old consumer can stop or update a newer request, an unmounted hook receives a state update, stopped HTTP work remains attached, or a JCEF error terminal resolves as success.
- Rollback or disable path: Revert this commit; there is no feature flag or data migration.

## Contributor declaration

AI assistance: Codex assisted with implementation, tests, and independent review. The maintainer inspected the source paths and ran the verification commands listed above.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,9 @@ jobs:
       - name: Run SQL execution stream contract
         run: yarn run test:sql-execution-stream
 
+      - name: Run SSE request ownership contract
+        run: yarn run test:sse-request
+
       - name: Run verification-code countdown lifecycle contract
         run: yarn run test:verification-code-countdown
 
--- a/chat2db-community-client/package.json
+++ b/chat2db-community-client/package.json
@@ -51,6 +51,7 @@
     "test:search-result-tab-selection": "tsx src/blocks/SearchResult/tabSelection.test.ts",
     "test:saved-console-tree-refresh": "tsx src/store/tree/savedConsoleTreeRefresh.test.ts",
     "test:saved-console-lifecycle": "tsx src/store/workspace/utils/savedConsoleLifecycle.test.ts && tsx src/store/workspace/utils/workspaceTabPersistence.test.ts",
+    "test:sse-request": "tsx src/components/SSERequest/requestOwnership.test.ts && tsx src/components/SSERequest/requestTransport.test.ts",
     "test:settings-layout": "tsx src/blocks/Setting/navigation.test.ts && tsx src/blocks/Setting/search.test.ts && tsx src/blocks/Setting/BaseSetting/model.test.ts && tsx src/blocks/Setting/settingsLayout.test.ts",
     "test:shortcut": "tsx src/constants/shortcut.test.ts && tsx src/utils/jcefZoom.test.ts",
     "test:sql-execution-log": "tsx src/service/sqlExecutionLog.test.ts",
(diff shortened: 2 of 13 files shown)


openai0229:
Closing this maintainer-owned replacement because the contributor enabled maintainer edits on the consolidated PR #2520. The complete request-scoped implementation, including AI create-table close handling, is now on #2520 at 7428b9e75c9ea8142a9eaeb49f646ae4c26ca906.

PR #2606, opened 2026-08-05
feat(mysql): visible/invisible index management for MySQL 8.0+

### LINKED ISSUE #2574: [MYSQL-OBJ-006] Manage visible and invisible MySQL indexes
## What needs to be done

The parser recognizes index visibility syntax, but index metadata and management do not show or change it. Display `VISIBLE` or `INVISIBLE` for ordinary indexes and let users preview and execute `ALTER TABLE ... ALTER INDEX ... VISIBLE|INVISIBLE`. The primary-key action must always be disabled with an explanation.

## Acceptance criteria

- [ ] Ordinary, unique, and composite indexes reload with the correct visibility state.
- [ ] Users can preview, execute, and reload visibility changes.
- [ ] The primary key is never offered as hideable.
- [ ] Before making an index invisible, the UI explains that the optimizer will not use it by default.
- [ ] MySQL 5.7 does not show the visibility action.

Support MySQL 8.0 and later, with InnoDB required test coverage. Primary keys cannot be invisible. Table-level `ALTER` is required. Do not analyze whether an index is safe to delete. There are no feature dependencies.

## Tests and test data

Add automated metadata, preview, visibility-change, reload, primary-key guard, privilege, and version-gating tests. Provide `script/test-fixtures/mysql/MYSQL-OBJ-006/init.sql`, `grants.sql`, `cleanup.sql`, and `README.md` containing ordinary, unique, composite, and primary indexes and verification queries for optimizer-visible state.

Aias00:
## Summary
- Add `visible` field to `TableIndex` domain model and read the `Visible` column from MySQL `SHOW INDEXES` (8.0+)
- Generate `INVISIBLE` keyword in index DDL via `MysqlIndexTypeEnum.buildIndexScript()`; add `buildAlterIndexVisibility()` for the `ALTER INDEX ... VISIBLE|INVISIBLE` path
- Frontend: add a "Visible" column to the index editor with a VISIBLE/INVISIBLE toggle, disabled for primary keys, gated to MySQL

Closes #2574

## Test plan
- [ ] MySQL 8.0+: open a table with indexes — verify VISIBLE/INVISIBLE state is shown
- [ ] Toggle an index from VISIBLE to INVISIBLE — verify SQL preview includes `ALTER INDEX ... INVISIBLE`
- [ ] Verify primary key row does not offer a visibility toggle
- [ ] MySQL 5.7: verify the visibility column does not appear
- [ ] `mvn -B compile` passes for the MySQL plugin module


openai0229:

P1: the visibility-only fast path is incomplete. `onlyIndexVisibilityChanged()` does not compare index method and other index attributes such as prefix length, so an edit that changes visibility together with another index property can be classified as visibility-only and silently omit the other DDL. Please compare the complete index definition (or remove this optimization), add regression tests for mixed visibility/method/prefix changes.

PR #2610, opened 2026-08-05
feat(mysql): index prefix length management

### LINKED ISSUE #2572: [MYSQL-OBJ-004] Manage MySQL index prefix lengths
## What needs to be done

MySQL index metadata can report `Sub_part`, but the index editor cannot display or preserve prefix lengths. Allow a prefix length per eligible index column, reload it accurately, and support create, delete, and drop-and-recreate modification with SQL preview. Validate column eligibility and every prefix-length limit that can be determined from the actual column type, character set, row format, storage engine, and server version before execution; preserve the MySQL error only for limits that cannot be predicted reliably.

## Acceptance criteria

- [ ] Single-column and composite prefix indexes can be created and reload with the correct length for each indexed column.
- [ ] Changing a prefix length previews the drop-and-recreate operation before execution.
- [ ] Ineligible column types, non-positive lengths, and limits determinable from the actual column, character set, row format, engine, and version metadata are rejected before submission.
- [ ] The original MySQL limit error is preserved only when the effective maximum cannot be calculated reliably before submission.
- [ ] Rebuilding an index warns about possible table locking and execution time.

Support MySQL 5.7 and 8.0; test InnoDB and MyISAM. Users need `INDEX` or equivalent table-level privileges. Eligible types include CHAR, VARCHAR, BINARY, VARBINARY, TEXT, and BLOB as allowed by MySQL. Do not calculate or recommend an optimal prefix length. There are no feature dependencies.

## Tests and test data

Add automated metadata, validation, preview, create, rebuild, delete, reload, server-limit, and privilege tests. Provide `script/test-fixtures/mysql/MYSQL-OBJ-004/init.sql`, `grants.sql`, `cleanup.sql`, and `README.md` with utf8mb4 VARCHAR, TEXT, binary, single-column, and composite index cases for both required engines.

Aias00:
## Summary
- Modify `MysqlIndexTypeEnum.buildIndexColumn()` to render `(prefix_length)` after the column name when `subPart` is present
- Uncomment the prefix length column in `IncludeCol` component, using `subPart` as the dataIndex, gated to MySQL only
- Add `subPart` field to `IIndexIncludeColumnItem` TypeScript type

The `subPart` field already existed on `TableIndexColumn` and was read from MySQL `SHOW INDEXES` (`Sub_part`), but the editor UI and DDL generation did not use it.

Closes #2572

## Test plan
- [ ] MySQL: open a table with prefix indexes — verify prefix lengths are displayed
- [ ] Create a new index with a prefix length — verify SQL preview includes `column_name(N)`
- [ ] Change a prefix length — verify SQL preview shows DROP + ADD with new prefix
- [ ] Verify prefix length reloads accurately after refresh
- [ ] Non-MySQL: verify the prefix length column does not appear
- [ ] `mvn -B compile` passes


openai0229:

P1: prefix length is not validated against the column type/length, and the implementation cannot reliably clear an existing prefix because an omitted/empty value is not distinguished from an unchanged value. This can generate invalid DDL or leave stale prefix metadata behind. Please add type-aware validation, explicit clear semantics, round-trip metadata tests.

PR #2614, opened 2026-08-05
feat(mysql): functional index expression support

### LINKED ISSUE #2573: [MYSQL-OBJ-005] Manage MySQL functional indexes
## What needs to be done

Functional-index expressions are not mapped from `SHOW INDEX`, and the index editor only accepts physical columns. Let users view, create, modify, and delete functional indexes, distinguish expression entries from column entries, generate valid double-parenthesized MySQL DDL, and avoid presenting MySQL's hidden generated columns as ordinary user-managed columns.

## Acceptance criteria

- [ ] An index such as `((lower(name)))` can be created and reloads with its complete expression.
- [ ] Valid indexes can combine expression entries and ordinary columns where MySQL permits it.
- [ ] Modifying an expression uses a previewed drop-and-recreate operation and reports the exact failure stage.
- [ ] MySQL-created hidden generated columns are represented without misleading users into editing them as normal columns.
- [ ] Unsupported MySQL versions do not show the functional-index entry.

Support MySQL 8.0.13 and later, with InnoDB required test coverage. Users need `INDEX` and/or `ALTER` as required by the operation. Preserve MySQL restrictions on deterministic expressions, LOB values, and prefixes, and warn about rebuild locking and duration. Do not recommend or generate expressions automatically. There are no feature dependencies.

## Tests and test data

Add automated metadata, create, composite-index, rebuild, delete, reload, failure, privilege, and version-gating tests. Provide `script/test-fixtures/mysql/MYSQL-OBJ-005/init.sql`, `grants.sql`, `cleanup.sql`, and `README.md` with a valid lowercase expression, a mixed expression/column index, an invalid nondeterministic function, and a MySQL 5.7 gating case.

Aias00:
## Summary
- Add `expression` field to `TableIndexColumn` domain model
- Read the `Expression` column from SHOW INDEXES (MySQL 8.0.13+, try/catch for 5.7)
- Render `((expression))` double-parenthesized syntax in `MysqlIndexTypeEnum.buildIndexColumn()`
- Frontend: display expression entries in IncludeCol with `((expression))` format

Closes #2573

## Test plan
- [ ] MySQL 8.0.13+: open a table with functional indexes — verify expressions are displayed
- [ ] Verify mixed index (expression + physical column) shows both entries
- [ ] Create a new functional index — verify SQL includes `((expression))` syntax
- [ ] MySQL 5.7: verify functional index options do not appear
- [ ] `mvn -B compile` passes


openai0229:

P1: functional-index expressions can acquire duplicated parentheses during DDL generation, and the table editor does not provide a complete reload/edit path for the expression. Please normalize expression ownership (builder vs caller), verify generated SQL against MySQL 8.0, add create/edit/drop/readback tests.

PR #2616, opened 2026-08-05
feat(mysql): trigger create and drop from object navigation

### LINKED ISSUE #2580: [MYSQL-OBJ-012] Create, modify, and delete MySQL triggers with recovery
## What needs to be done

Triggers can currently be listed and their DDL opened, but there is no complete create, replace-with-recovery, delete, and refresh workflow. Let users choose the target table, `BEFORE`/`AFTER`, `INSERT`/`UPDATE`/`DELETE`, body, Definer, and supported same-event order. For modification, capture the original definition, preview the drop and create phases, and attempt to restore the old trigger if recreation fails.

## Acceptance criteria

- [ ] All six timing/event combinations can be created, execute correctly, and reload with their complete definition.
- [ ] Supported `FOLLOWS` or `PRECEDES` order is retained after refresh.
- [ ] Modification previews both phases and reports the failed phase and restoration result.
- [ ] If restoration succeeds, the previous trigger remains executable; restoration failure is reported explicitly.
- [ ] Successful deletion removes the trigger from navigation and table details after confirmation.

Support MySQL 5.7 and 8.0 with any target-table engine; show `FOLLOWS`/`PRECEDES` only when supported. Users need `TRIGGER` on the target table, and non-current Definers remain subject to MySQL permissions. Delete and replacement require confirmation. Do not add trigger debugging, side-effect analysis, or cross-server triggers. Existing routine replacement recovery may be reused as behavior.

## Tests and test data

Add automated create, execution, ordering, preview, replacement, restoration-success, restoration-failure, delete, refresh, Definer, and privilege tests. Provide `script/test-fixtures/mysql/MYSQL-OBJ-012/init.sql`, `grants.sql`, `cleanup.sql`, and `README.md` with all six combinations, OLD/NEW references, multiple ordered triggers, complex bodies, and controlled replacement failures.

Aias00:
## Summary
- Add `drop()` method to `IDbTriggerService`, implemented in `DbTriggerServiceImpl`
- Add `POST /api/rdb/trigger/delete` endpoint to `DbTriggerController`
- Add `deleteTrigger` frontend service method to `sql.ts`
- Add `CreateTrigger` and `DropTrigger` tree menu operations with handlers

Closes #2580

## Test plan
- [ ] Right-click "Triggers" → "Create trigger" — verify editor opens
- [ ] Right-click existing trigger → "Drop trigger" — verify confirmation
- [ ] Verify all 6 timing/event combinations work
- [ ] Verify FOLLOWS/PRECEDES ordering on MySQL 5.7+
- [ ] `mvn -B compile` passes


openai0229:

P1: trigger deletion has the same request mismatch as the create/drop flow: the client sends `tableName`, while the backend expects `triggerName`. The delete action cannot reliably identify the trigger. Please align the payload contract, add an API test for the real client request, and fix the failing CodeQL check

PR #2620, opened 2026-08-06
feat(mysql): JSON invalid-submission protection in value processor

### LINKED ISSUE #2583: [MYSQL-DATA-002] Prevent invalid MySQL JSON updates and preserve scalar and NULL semantics
## What needs to be done

Make the existing JSON editor handle every legal JSON root value: object, array, string, number, boolean, and JSON `null`. Show JSON syntax errors with line and column information, prevent invalid content from entering a pending DML update, provide an explicit SQL NULL action that is separate from JSON `null`, and reload the MySQL-normalized value after saving.

## Acceptance criteria

- [ ] All six legal JSON root types are recognized, edited, saved, and read back as JSON.
- [ ] Invalid commas, quotes, and truncated input cannot enter pending DML, and the first error is shown with its line and column.
- [ ] JSON `null`, the JSON string `""`, and SQL NULL remain three distinct values after saving and reloading.
- [ ] Numbers, booleans, and their string equivalents are not coerced into one another.
- [ ] Formatting and minification preserve JSON meaning, and cancelling an edit leaves the original value unchanged.

Support native JSON columns in MySQL 5.7.8+ and 8.0. Compatible databases without native JSON support must continue to use the text editor. Reuse the existing safe result-set DML path and permissions. Error messages must not expose unrelated row data. JSON Schema validation, JSON path query building, and document merge or diff are out of scope.

## Tests and test data

Add automated tests for every legal root type, Unicode, deep nesting, malformed commas and quotes, truncated input, JSON `null`, an empty JSON string, SQL NULL, large content, cancellation, and save-and-reload behavior. Provide repeatable `init.sql`, `grants.sql`, `cleanup.sql`, and `README.md` files, plus a JSON fixture file containing valid and invalid editor inputs. The README must document setup, exact test commands, expected results, and cleanup.

Aias00:
## Summary
- Add `getSqlValueString` override in `MysqlValueProcessor` that validates JSON column values using fastjson2 `JSON.parse()` before generating SQL
- Throws `BusinessException` with parse error message for invalid JSON, preventing invalid content from entering DML updates
- Skips validation for null values (SQL NULL), function values (now(), default), and non-JSON types

Closes #2583

## Test plan
- [ ] Edit a JSON cell with valid JSON — verify it saves correctly
- [ ] Edit a JSON cell with INVALID JSON — verify submission is rejected
- [ ] Verify JSON null, empty string, and SQL NULL are distinct after save/reload
- [ ] Verify numbers and booleans are not coerced into strings
- [ ] `mvn -B compile` passes
- [ ] `eslint` passes


openai0229:

P1: the JSON validation path accepts only the first Jackson JSON value and does not require end-of-input. Empty input and valid JSON followed by trailing content can therefore be accepted incorrectly. Please use a parser that verifies exactly one value plus EOF, define scalar/NULL semantics, add malformed/trailing-token tests.

PR #2624, opened 2026-08-06
feat(mysql): role lifecycle management (MySQL 8.0+)

### LINKED ISSUE #2591: [MYSQL-SEC-004] Add MySQL Role lifecycle, grants, defaults, and active-role state
## What needs to be done

Add MySQL Role management for creating and deleting Roles, granting and revoking Roles to users or other Roles, `WITH ADMIN OPTION`, nested Role relationships, and default Roles using ALL, NONE, or a selected set. Show Roles in the object tree and SQL completion. Clearly distinguish Roles that are granted, enabled by default, and active in the current connection. Read `CURRENT_ROLE()` and, once MYSQL-SQL-001 provides a fixed Console session, allow `SET ROLE` for that Console only.

## Acceptance criteria

- [ ] A Role can be created, granted database privileges, granted to a user, configured as default, and observed on a new connection.
- [ ] Roles can be granted to and revoked from users or other Roles, including `WITH ADMIN OPTION`, without losing relationship state after refresh.
- [ ] Deleting or revoking a Role removes its relationships and SQL-completion candidate after refresh.
- [ ] Cyclic Role grants are rejected before execution or surfaced with the server's clear rejection.
- [ ] Granted, default-enabled, and currently active Roles are displayed as different states.
- [ ] `SET ROLE` affects only the bound Console session, and MySQL 5.7 exposes no Role-management entry points.
- [ ] Role deletion requires entering the Role name for confirmation.

Support MySQL 8.0+ only and hide the feature on MySQL 5.7. Enforce `CREATE ROLE`, `DROP ROLE`, `ROLE_ADMIN`, applicable `GRANT OPTION`, equivalent server permissions, and `SYSTEM_USER` restrictions. Role lifecycle and grants may be delivered independently, but current-session `SET ROLE` depends on MYSQL-SQL-001. MySQL 5.7 Role emulation, cross-server Roles, and automatic least-privilege design are out of scope.

## Tests and test data

Add automated tests for tree and completion refresh, create/delete, user and Role recipients, nested and cyclic grants, revoke, ADMIN OPTION, default ALL/NONE/selected sets, `CURRENT_ROLE()`, Console-scoped `SET ROLE`, MySQL 5.7 gating, deletion confirmation, and insufficient permissions. Provide repeatable `init.sql`, `grants.sql`, `cleanup.sql`, and `README.md` files with isolated users, Roles, nested relationships, and representative privileges. The README must document setup, exact test commands, version prerequisites, expected effective-role states, and cleanup.

Aias00:
## Summary
- Add `CREATE_ROLE`, `DROP_ROLE`, `GRANT_ROLE`, `REVOKE_ROLE`, `SET_DEFAULT_ROLE` action types
- Add `roleName`, `roleList`, `withAdminOption`, `defaultRoleMode` fields
- Add DDL cases in `MysqlAccountSqlBuilder`:
  - `CREATE ROLE` / `DROP ROLE IF EXISTS`
  - `GRANT role TO user [WITH ADMIN OPTION]`
  - `REVOKE role FROM user`
  - `SET DEFAULT ROLE ALL|NONE|role1,role2 TO user`
- Add 5 role action types and role fields to frontend types

Closes #2591

## Test plan
- [ ] Create a role — verify SQL shows `CREATE ROLE ...`
- [ ] Grant role to user with WITH ADMIN OPTION — verify SQL
- [ ] Set default role to ALL/NONE/selected — verify SQL
- [ ] Revoke a role — verify SQL shows `REVOKE role FROM user`
- [ ] Drop a role — verify SQL shows `DROP ROLE IF EXISTS ...`
- [ ] Verify nested role grants display correctly
- [ ] On MySQL 5.7, verify role management is not shown
- [ ] `mvn -B compile` passes
- [ ] `eslint` passes


openai0229:

P1: role grants are generated with backtick-quoted identifiers, but MySQL role accounts use the `'role'@'host'` account form. The model has no host component, so composite role accounts cannot be represented. There is also no MySQL 8.0 gate, role tree/completion integration, or active/default-role readback. Please correct the account grammar and complete the role lifecycle workflow in the shared account stack before rebasing.

PR #2628, opened 2026-08-06
feat(mysql): structured TLS configuration for MySQL connections (#2568)

### LINKED ISSUE #2568: [MYSQL-CONN-001] Add structured MySQL TLS configuration and diagnostics
## What needs to be done

MySQL TLS currently requires users to enter Connector/J parameters manually. Add structured connection settings for `REQUIRED`, `VERIFY_CA`, and `VERIFY_IDENTITY`, with CA PEM, client certificate PEM, PKCS#8 private key, and optional JKS/PKCS12 key/trust stores. Web and Desktop must upload file contents rather than retain local paths. Show a redacted preview of the equivalent JDBC settings, diagnose certificate, hostname, and private-key failures, persist sensitive material with the same protection as datasource passwords, and remove obsolete material when a connection is updated or deleted.

## Acceptance criteria

- [ ] Users can configure and test one-way TLS and mutual TLS without manually composing JDBC parameters.
- [ ] `REQUIRED`, `VERIFY_CA`, and `VERIFY_IDENTITY` work with Connector/J 5.1.47 and 8.0.30 using the correct version-specific parameters.
- [ ] Saved settings reopen without loss, while private keys, store passwords, and sensitive URL parameters never appear in previews, logs, exports, or errors.
- [ ] Expired certificates, untrusted CAs, hostname mismatches, and incorrect private-key passwords produce distinct, actionable errors.
- [ ] Replacing or deleting TLS material makes the previous material unavailable.

Support MySQL 5.7 and 8.0 with either supported Connector/J version and any storage engine. No database administration privilege is required. Connector/J 5.1 compatibility must use its `useSSL`, `requireSSL`, and `verifyServerCertificate` family where appropriate; Connector/J 8 uses `sslMode`. Do not issue certificates, change server TLS configuration, invoke the `mysql` CLI, depend on an operating-system certificate store, or provide a standalone certificate-conversion tool.

## Tests and test data

Add automated connection, persistence, redaction, replacement, and deletion tests covering Web and Desktop across both Connector/J versions. Provide `script/test-fixtures/mysql/MYSQL-CONN-001/init.sql`, `grants.sql`, `cleanup.sql`, and `README.md`, plus a reproducible `compose.yaml` and certificate fixtures for a trusted CA, an untrusted CA, an expired certificate, a wrong hostname, mutual TLS, and an incorrect private-key password. The README must document certificate generation or provenance and exact commands for every connection matrix case.

Aias00:
## Summary

Closes #2568. Adds a structured TLS configuration path for MySQL so users can configure one-way and mutual TLS (CA / client cert / private key, or a pre-built keystore) instead of relying on raw URL params. This PR lands the **backend** plus a bug fix and tests; the frontend UI is tracked as follow-up.

## Backend infrastructure

- **`SSLInfo`** — structured TLS fields (`tlsMode`, `caPem`, `clientCertPem`, `clientPrivateKeyPem`, `clientKeyPassword`, `keyStoreType/Bytes/Password`); `@ToString` excludes the secret fields.
- **`MySqlTlsMode`** enum — `DISABLED` / `REQUIRED` / `VERIFY_CA` / `VERIFY_IDENTITY`.
- **`MySqlTlsTranslator`** — translates `SSLInfo` into Connector/J connection properties, handling **8.x** (`sslMode` + PEM/keystore `data:` URLs) vs **5.1.x** (`useSSL`/`requireSSL`/`verifyServerCertificate`). Properties are used (not URL params) so they survive the SSH-tunnel URL rewrite.
- **Encryption** — `LocalWorkspaceStorage` encrypts secret TLS fields on create and preserves/re-encrypts them on update (mirrors the password preserve-if-blank rule); public material stays cleartext. `DbWorkspaceDataSourceServiceImpl` decrypts for LOCAL (AES) and cloud (RSA token) storage and restores saved TLS material on test-from-saved.
- Wired into `DefaultDBManager` (live connect) and `DbDataSourceServiceImpl` (test connect). Persistence is automatic via MapStruct (`ssl` is a JSON column like `ssh` — **no schema migration**).

## Bug fix

`JdbcDriverManager.tryConnectionAgain` unconditionally injected `useSSL=false` on retry, which clobbered Connector/J **8.x** TLS — the translator sets `sslMode`, not `useSSL`, so a failed v8 TLS connection was retried with `useSSL=false`, conflicting with `sslMode` and risking silently disabling the user's TLS. Wired `MySqlTlsTranslator.hasExplicitTlsIntent()` (broadened to `Map<?, ?>` so it inspects both the driver `Map` and the retry-path `Properties`) so the retry skips `useSSL=false` when any TLS intent is present.

## Tests

All green. Run with `-Dmaven.test.skip=false` (the BOM sets `maven.test.skip=true` by default).

| Module | Tests |
| --- | --- |
| `chat2db-community-spi` | 142 pass (incl. 27 new in `MySqlTlsTranslatorTest`) |
| `chat2db-community-storage` | 37 pass (incl. 10 new in `LocalWorkspaceStorageSslMergeTest`) |
| `chat2db-community-domain-core` | 125 pass (incl. 4 new in `DbWorkspaceDataSourcePasswordSemanticsTest`) |

Coverage:
- **`MySqlTlsTranslatorTest`** — v8/v5 property sets per mode, PEM `data:` URLs, keystore override, blank-skipping, version detection, `hasExplicitTlsIntent` truth table (including the `Properties` retry path).
- **`LocalWorkspaceStorageSslMergeTest`** — create-encrypt, update preserve/re-encrypt semantics, public-field clearing, null handling. (AES-GCM is non-deterministic, so ciphertext is verified by round-tripping through `decrypt`, not by string equality.)
- **`DbWorkspaceDataSourcePasswordSemanticsTest`** — `preConnect` TLS recovery: restores saved TLS when the request omits it, keeps resubmitted TLS intact, treats whitespace-only as blank.

## Not included

Frontend TLS UI does not exist yet — users cannot currently configure TLS from the connection form. This is the follow-up to make the feature end-to-end usable.


openai0229 (changes requested):
Request changes. The current backend does not yet satisfy the issue's connection/persistence/security acceptance criteria.

1. [P1] TLS material cannot be consumed by Connector/J 5.1.47 or 8.0.30 - `chat2db-community-server/chat2db-community-spi/src/main/java/ai/chat2db/spi/sql/MySqlTlsTranslator.java:139-181`. The translator emits `data:...` URLs, but both driver versions call `new URL(url)` and then `openStream()`; the JDK has no `data:` URL handler. It also emits `trustCertificateKeyStoreType=PEM` / `clientCertificateKeyStoreType=PEM`, while both drivers load the value through `KeyStore.getInstance(type)` and only document JKS/PKCS12. Any CA PEM, client PEM, or in-memory JKS/PKCS12 upload therefore fails before a TLS handshake. The tests only assert property strings and do not exercise either driver. Please use a supported in-memory/temporary-file strategy (and cover both actual driver versions).

2. [P1] Sensitive TLS fields are returned by normal datasource APIs - `chat2db-community-server/chat2db-community-domain/chat2db-community-domain-core/src/main/java/ai/chat2db/community/domain/core/impl/db/DbWorkspaceDataSourceServiceImpl.java:65-69,219-260`. `queryDisplayDataSourceById(id, false)` is used by `GET /api/connection/datasource` and create/update responses, but `decryptSensitiveFields` always decrypts `clientPrivateKeyPem`, `clientKeyPassword`, `keyStoreBytes`, and `keyStorePassword` regardless of `requestPassword=false`. `DataSourceWebConverter.storage2response` maps the object directly, so the private key and passwords are sent to the frontend. Keep secrets unavailable in display/export responses and decrypt only inside the connection/test path.

3. [P1] Community export leaks encrypted TLS secrets and cannot round-trip them - `chat2db-community-server/chat2db-community-domain/chat2db-community-domain-core/src/main/java/ai/chat2db/community/domain/core/impl/db/DbWorkspaceDataSourceServiceImpl.java:184-192`. The community branch calls `queryDataSourceById(..., true)`, which returns the at-rest ciphertext from `LocalWorkspaceStorage`, then returns it unchanged in `ssl`. This violates the requirement that private keys/store passwords never appear in exports; importing that JSON would also encrypt the ciphertext again. Redact secret TLS fields from export (or define an explicit, separately protected export format) and add an export/redaction test.

4. [P1] Clearing TLS does not remove the previous private material - `chat2db-community-server/chat2db-community-storage/src/main/java/ai/chat2db/community/storage/LocalWorkspaceStorage.java:277-306` and `DbWorkspaceDataSourceServiceImpl.java:84-87`. A null/blank `SSLInfo` is treated as preserve old value, and every blank secret field preserves the previous ciphertext. This makes it impossible to disable TLS and delete an old key/password as required by the issue; `preConnect` also restores the saved TLS config when the request explicitly sends a blank config. Distinguish omitted fields from explicit deletion and clear the stored ciphertext when the user removes TLS material.

The PR also does not include the required Web/Desktop TLS UI, trust-store fields, actionable certificate/hostname/private-key diagnostics, or the issue's integration fixtures; please either add them or split this into a clearly scoped backend infrastructure PR with the acceptance criteria tracked separately.

openai0229:
Follow-up review after the latest head: the four P1 implementation issues from review `4970742360` are addressed in the current tree. Connector/J 5.1/8.0 now receive file-backed JKS/PKCS12 stores with direct driver-loading tests, normal responses and exports redact TLS secrets, and explicit clearing removes stored private material. The conflict with `main` is also resolved.

The PR is still not complete against #2568: there is no Community Web/Desktop TLS configuration workflow, no actionable certificate/hostname/private-key diagnostics, and no real MySQL TLS handshake fixture proving the end-to-end connection path. Please complete that vertical workflow and let the current CI finish before merge; the CHANGES_REQUESTED state remains for these acceptance gaps.

Aias00:
@openai0229 The requested TLS changes have been addressed on current head `6d778209d`: Connector/J 5.1/8 driver-readable material handling, API/export secret redaction, explicit TLS clearing, Web/Desktop UI, actionable diagnostics, and real TLS fixtures/tests. All current-head checks, including Java CodeQL and the independent CodeQL alert gate, are successful. Please re-review when available.

Issue #2586, opened 2026-08-05
[MYSQL-SQL-001] Add manual transaction mode and session protection to MySQL consoles


openai0229:
## What needs to be done

Add a manual transaction mode scoped to one SQL Console. When auto-commit is off, repeated executions in that Console must use one isolated JDBC connection and expose transaction state plus Commit and Rollback controls. Protect uncommitted work when the Console closes, its connection changes, the network disconnects, or the application exits, and always release the bound connection safely.

## Acceptance criteria

- [ ] Two DML executions in one Console remain invisible to another connection until Commit, become visible after Commit, and disappear after Rollback.
- [ ] Different Consoles never share a transaction or bound connection.
- [ ] Closing a Console with uncommitted changes requires the user to Commit, Rollback, or cancel closing.
- [ ] Cancellation, timeout, execution failure, and network loss do not leak a connection and report whether the transaction was rolled back or its outcome is unknown.
- [ ] DDL that can implicitly commit produces a warning before execution.
- [ ] Transaction and auto-commit state remain clear throughout the Console workflow.

## Compatibility and constraints

Support MySQL 5.7 and 8.0, with InnoDB covered as the transactional baseline. Clearly state that non-transactional engines such as MyISAM are not protected by Rollback. Do not add database privileges, reuse a bound connection for another request, or expose credentials and transaction SQL outside existing log-redaction rules. Distributed transactions, XA management, a savepoint GUI, and automatic replay of an entire transaction are out of scope.

## Tests and test data

Add automated integration tests for Commit, Rollback, visibility from a second connection, isolation between Consoles, implicit DDL commit, non-transactional tables, cancellation, timeout, disconnect, all three close choices, application shutdown, and connection-pool return. Provide repeatable `init.sql`, `grants.sql`, `cleanup.sql`, and `README.md` files with separate test accounts and InnoDB/MyISAM tables. The README must document setup, exact concurrent test steps or commands, expected visibility, failure behavior, and cleanup.

PR #2630, opened 2026-08-06
feat(mysql): manual transaction mode with console-scoped session protection (#2586)

Aias00:
## Summary

Implements #2586 [MYSQL-SQL-001]: a manual (auto-commit off) transaction mode for SQL Consoles. A user can run multiple statements inside one transaction and choose to commit or roll back, instead of every statement auto-committing immediately. The bound connection is pinned to the console across executions and released cleanly on commit/rollback, console close, connection switch, datasource edit, and shutdown.

## Backend

- **`ConsoleTransactionRegistry`** (new, spi) pins one isolated JDBC connection per console (`ConnectInfo.consoleOwn=true`). `ConnectionPool.close` and `Chat2DBContext.removeContext` skip the pool return while a transaction is open, so the bound connection is reused across executions and its lifecycle is owned by the registry.
- Commit / rollback / release restore auto-commit and return the connection to the pool; `release` rolls back first. Bound connections are swept on datasource edit (`ConnectionPool.removeConnection` → `releaseByDataSourceId`), app shutdown (`ApplicationLifecycleAdapter` `@PreDestroy` → `releaseAll`), and after a 30-min idle timeout by a daemon thread (safety net for abandoned consoles).
- **`DbConnectionContextServiceImpl.bind`** grafts the bound connection into the fresh `ConnectInfo` so executions reuse it (hits `ConnectionPool.tryGetExistingConnection` fast path); `rebindCurrentDatabase` releases the bound transaction first because a catalog switch cannot survive.
- **`DefaultSQLExecutor`** warns when an implicit-commit DDL (CREATE/ALTER/DROP/TRUNCATE/RENAME/GRANT/REVOKE/FLUSH/…) runs inside an open transaction; forwarded via `ISqlExecutionStatementListener.onImplicitCommitWarning` and streamed to the client. `SqlExecutionJob` carries the console `inTransaction` flag in terminal events so the toolbar stays in sync (including after cancellation, which leaves an open transaction intact by design).
- New endpoints `POST /api/rdb/transaction/{begin,commit,rollback,state,release}` (`DbTransactionController`), bound by `@ConnectionInfoAspect`. `TransactionStateResponse` reports `inTransaction`, `mode`, `outcome` (incl. `UNKNOWN` when a commit/rollback failed and the connection had to be discarded), and `lastError`.

## Frontend

- `service/transaction.ts` for the five transaction endpoints.
- Console store holds a **non-persisted** `transactionStateMap` keyed by `consoleId` (`TransactionState{mode, inTransaction, lastOutcome, lastError}`) — intentionally excluded from persistence so a reload never claims a transaction is open when the server registry (which does not survive restart) has none.
- `useSqlExecutor.ensureTransactionBegun` triggers `begin` lazily before the first execution in manual mode; on failure it falls back to auto-commit so a transient begin error never blocks the user.
- `SQLEditorWithOperation` / `OperationLine` render the auto-commit toggle plus commit/rollback controls (shown in manual mode); `SQLExecute` surfaces DDL implicit-commit warnings.
- `utils/transactionSession.confirmAndReleaseTransaction` prompts commit/rollback/cancel on console close and connection switch, wired into `WorkspaceTabs` close paths and `deleteActiveWorkspaceTab`.

## Tests

`MysqlConsoleTransactionIT` exercises the registry against a real MySQL 8.0 via Testcontainers — commit visibility, rollback invisibility, console isolation (two consoles don't see each other's uncommitted work), MyISAM non-protection, and release without connection leak. Test tables are truncated per test for isolation; bound-connection identity is captured before commit (commit nulls the bound connection). The mysql plugin gains Testcontainers + `mysql-connector-j` (test scope) dependencies.

## Verification

- Backend compiles: `chat2db-community-spi`, `-domain-core`, `-web` (via `-am`), mysql plugin test-compile.
- `MysqlConsoleTransactionIT` — **5/5 passing** (real MySQL 8.0 via Testcontainers).
- Client: `pnpm test:i18n` passes; eslint clean on all changed files; no new `tsc` errors (the remaining `tsc` errors are pre-existing baseline noise in `APNGPlayer`, `BI`, `AI` blocks, unrelated to this change).

## Notes

- This is the first module to use Testcontainers, so the Testcontainers BOM is imported locally in the mysql plugin pom.
- MySQL JDBC driver is not a project dependency (Chat2DB loads production drivers dynamically at runtime from an external directory); the IT declares `mysql-connector-j` as a test-scoped dependency.


openai0229:

The latest Backend and CodeQL checks fail because the branch does not compile against the current `IDbConnectionContextService` contract (`isInTransaction(Long)` is missing in the test double). Beyond CI, `commit/rollback/release` are not serialized per console, `release` races with in-flight operations, and endpoints bind by `consoleId` without validating data-source identity or user ownership. Please fix the contract and concurrency/authorization model, add HTTP-level and concurrent tests for MySQL 5.7/8.0.

PR #2632, opened 2026-08-06
fix(import): gracefully handle undecryptable DBeaver credentials, support optional master password

### LINKED ISSUE #2631: [Bug] DBeaver .dbp import aborts with BadPaddingException when credentials are encrypted with a non-default key (e.g. master password set)
## Description

Importing a `.dbp` file exported from DBeaver fails with `javax.crypto.BadPaddingException` and the whole import aborts, when the DBeaver workspace encrypted its `credentials-config.json` with a key other than DBeaver's default local key (most commonly when a **master password** is set in DBeaver).

## Steps to reproduce

1. In DBeaver, set a master password (Database -> Security -> Master password)
2. Export the project as `.dbp` (File -> Export -> DBeaver -> Projects)
3. In Chat2DB, import the `.dbp` (Import connection -> DBeaver)
4. Import fails with the error below

## Expected behavior

The import dialog itself states: *"We cannot import connection passwords. After the import is complete, please re-enter the connection password"* — so undecryptable credentials should be **skipped** and connections imported without passwords, not abort the whole import.

## Actual behavior

```
javax.crypto.BadPaddingException: Given final block not properly padded. Such issues can arise if a bad key is used during decryption.
    at ai.chat2db.community.domain.core.impl.ncx.dbeaver.DefaultValueEncryptor.decryptValue(DefaultValueEncryptor.java:77)
    at ai.chat2db.community.domain.core.impl.ncx.TaskNcxImportServiceImpl.dbpUploadFile(TaskNcxImportServiceImpl.java:169)
    at ai.chat2db.community.web.api.controller.DbConverterController.edbpUploadFile(DbConverterController.java:96)
```

A related crash also occurs (NullPointerException) when a connection has no entry in the credentials file (e.g. "save password" was disabled for it), because `TaskNcxImportServiceImpl.java` dereferences `userInfo` without a null check.

## Environment

- Chat2DB Community 5.3.3 desktop (arm64)
- DBeaver with master password set

## Notes

`DefaultValueEncryptor#makeSecretKeyFromPassword` (key = first 16 bytes of the master password, UTF-8) already exists in the codebase but is never used — it enables decrypting credentials when the user provides their DBeaver master password.

tsonglew:
## Summary

Fixes two crashes that abort the whole **DBeaver .dbp connection import**, adds optional support for decrypting credentials with the DBeaver **master password**, and hardens the archive extraction that the import performs:

1. **BadPaddingException aborts the import** (`TaskNcxImportServiceImpl`): when `credentials-config.json` in the .dbp was encrypted with a key other than DBeaver's default local key (e.g. a master password is set in DBeaver), `DefaultValueEncryptor.decryptValue` throws `javax.crypto.BadPaddingException` and the entire import fails. The import dialog itself states *"We cannot import connection passwords. After the import is complete, please re-enter the connection password"* — so undecryptable credentials are now **skipped with a warning** and connections are imported without passwords.

2. **NullPointerException** when a connection has no entry in the credentials file (e.g. "save password" was disabled for it): the credential lookup is now null-guarded and skipped.

3. **Optional master password** (new feature): when provided, credentials are decrypted with the key derived from the DBeaver master password — `DefaultValueEncryptor#makeSecretKeyFromPassword` (UTF-8 first 16 bytes), the same derivation as DBeaver's `DefaultValueEncryptor`. This method already existed in the codebase but was never used.

4. **Archive extraction is contained and bounded** (added in review follow-up): the project and resource names of a `.dbp` come from the archive's own `meta.xml`, so every destination is now resolved inside a directory created for that import alone, cleanup removes exactly that directory, and extraction is bounded in entry count and total size. Details in *Review follow-up* below.

## Changes

**Backend**

- `DbeaverCredentialsResolver` (new, `impl/ncx/dbeaver/`): reads `credentials-config.json`, trying (1) the default local key, then (2) the master-password-derived key. Failure to decrypt, a missing file, an unreadable file, and an empty or non-object document all resolve to "credentials unavailable" so the import continues without passwords. The parsed document is **cached per credential file, including the unavailable outcome**, so each file is read and decrypted once per import instead of once per connection. Instances are request-scoped; the master password and decrypted credentials are never persisted, logged, or cached statically.
- `DbeaverArchiveExtraction` (new, `impl/ncx/dbeaver/`): extracts one archive into a directory created for that import alone (`dbp-import-*` below the Chat2DB base path). A project or resource name must be a single path segment whose resolved target stays directly inside the directory it belongs to; anything else is rejected with a localized message (`connection.import.dbeaver.unsafeArchive`). Extraction is bounded by 10 000 entries and 256 MiB, charged with the bytes actually copied rather than the size the archive declares (`connection.import.dbeaver.archiveTooLarge`).
- `TaskNcxImportServiceImpl.java`: `dbpUploadFile(file, masterPassword)`; credential decryption delegated to the resolver once per project; `applyDbeaverCredentials` null-guards a missing connection entry and tolerates a malformed entry; the uploaded archive and **the private extraction directory of this import** are removed in a `finally` block so nothing survives a failed import; the `data-sources.json` stream is closed.
- `ITaskNcxImportService.java`: signature updated with the optional `masterPassword` parameter and its lifetime documented, plus a `dbpUploadFile(File)` default overload so callers without a password keep their call shape.
- `DbConverterController.java`: `POST /api/converter/dbp/upload` accepts an optional `masterPassword` form parameter.
- Message bundles: `connection.import.dbeaver.unsafeArchive` and `connection.import.dbeaver.archiveTooLarge` in all five locales and the default bundle.

**Frontend**

- `ImportConnection/index.tsx`: optional master-password input (password field) shown for the DBeaver import type; sent as a form field when non-empty. The field is controlled and the value is dropped when the dialog closes and after a settled request, so a password typed for an earlier import is never sent silently with the next one.
- `utils/sensitiveParams.ts` (new): both transports redact password-like fields from the request parameters they attach to an error notification, so the password cannot be copied out of the error dialog or printed by the desktop IPC log.
- i18n: `connection.import.dbeaver.masterPassword.placeholder` for all five locales (en-US, zh-CN, es-ES, ja-JP, ko-KR) plus refreshed `scripts/i18n-source-hashes.json`, which `yarn test:i18n` gates.

**Transport and privacy decision (Community web mode)**

`SECURITY.md` gains a *"Credentials supplied for an import"* section recording the explicit decision: the master password is request-scoped input, used only to decrypt that archive's credential files, never written to Chat2DB storage and never returned to a client; both clients remove password-like fields from the request details attached to an error notification; the password travels in the request body, which the existing `/api/**` access log already records under the positional masking every other request body receives; the archive is extracted into a private per-import directory that is always removed, whose names cannot leave it and whose size is bounded; supplying the password is optional. It crosses the HTTP boundary in the same request form as the connection passwords the datasource endpoints already accept, so it inherits the existing loopback-only deployment boundary rather than adding a new one — a network-reachable Community deployment is unsupported and would need HTTPS. The same boundary is referenced from the controller javadoc.

## Review follow-up

The review asked for two filesystem safety fixes in the archive extraction path, and the security report ([#2632 comment](https://github.com/OtterMind/Chat2DB/pull/2632#issuecomment-5598571600)) raised Zip Slip and Zip Bomb. All three are addressed by extracting into a directory that belongs to one import:

- **Zip Slip**: destinations were built by concatenating archive-controlled names, so a crafted `.dbp` could write outside the base path and the `finally` cleanup could then delete an attacker-selected path. Names are now validated (single path segment, no separators, no `.`/`..`) *and* every resolved target is normalised and checked to be a direct child of the directory it belongs to. Regression tests cover project names, resource names, nested and absolute names.
- **Deleting a directory the import did not create**: cleanup deleted `basePath/<project name>` taken from the archive, which removed a pre-existing directory when an archive reused a name — including Chat2DB's own state directories that live next to the extraction target. Cleanup now removes only the directory this import created.
- **Zip Bomb**: extraction had no size or entry bound, and `spring.servlet.multipart.max-file-size` is `-1`. Extraction now stops at an entry count and a byte budget charged with the bytes actually written.
- **Master password in error details**: `requestParams` was serialised whole into the error notification (`SystemErrorMessage` copies it to the clipboard), which contradicted the privacy statement. Password-like fields are now redacted before those parameters are reported, and `SECURITY.md` was corrected accordingly.

## Tests

New focused tests in `chat2db-community-domain-core` (24 tests):

- `DbeaverCredentialsResolverTest` (8): default-key success, master-password success, missing/blank/wrong master password, missing file, malformed files (plaintext, empty, truncated, decryptable-but-not-JSON), decrypt-once-per-file caching, cached unavailable outcome, and per-project isolation.
- `DbeaverArchiveExtractionTest` (6): declared resources land inside the extraction directory; project and resource names with traversal segments, and names that are not a single segment, are rejected without writing outside the directory; archives that exceed the entry budget and the byte budget are rejected.
- `TaskNcxImportServiceDbeaverImportTest` (10): end-to-end .dbp import with the default key and with the master password, import without passwords when credentials cannot be decrypted, absent credential file, missing and malformed credential entries, multi-project cleanup on both the success and failure paths, rejection of archives whose project or resource name escapes the extraction directory, and preservation of a pre-existing directory that shares a project name.
- `sensitiveParams.test.ts` (frontend, run by `yarn test:sensitive-params`): the DBeaver import parameters keep the password out of error details, nested credentials are redacted, absent values stay absent, and the sent parameters are not mutated.

Caching and cleanup are mutation-checked: removing the cache fails `decryptsEachCredentialFileOnlyOnce`, and moving cleanup off the `finally` path fails the cleanup tests.

## Verification

- `mvn test` on `chat2db-community-domain-core` with tests explicitly enabled (stale Surefire include and failure-ignore overridden): **317 tests, 0 failures, 0 errors**.
- `mvn test` on `chat2db-community-web`: **84 tests, 0 failures, 0 errors**.
- `mvn install` of `chat2db-community-domain-core` and `chat2db-community-web` with their dependencies: **BUILD SUCCESS**.
- `yarn lint` (eslint + stylelint): clean. `yarn test:i18n`: clean. `yarn test:sensitive-params`: clean.
- `yarn build:web:community --app_version=0.0.0` (prebuild test battery, which now includes `yarn test:sensitive-params`, webpack, and production bundle verification): **success**.
- The five new message keys were loaded with `java.util.Properties` from all six bundles to confirm they parse and resolve.
- End-to-end locally: a real .dbp with 30 connections exported from DBeaver (master password set) previously failed with the exact BadPaddingException from the issue; after the fix the credentials decrypt with the master password and all 30 connections import, including users/passwords.

Merged current `main` (through `c3e73f38`), which also brings upstream's `data-sources.json` stream-close fix (#2546) — this branch already contained the same close, so the merge keeps one implementation.

## Related
- Fixes #2631


openai0229:
Follow-up review after head `bf46f86`: the per-project credential cache, default-key/master-password tests, malformed/missing-entry coverage, cleanup tests, and loopback/HTTPS transport documentation are now present.

Changes are still requested for two safety issues in the archive extraction path:

1. `importDbeaverConfig()` builds destination paths from archive-controlled project/resource names (`TaskNcxImportServiceImpl.java:337-355`) without canonicalizing or enforcing a destination root. A crafted `.dbp` can use `../` entries to write outside the Chat2DB base path (Zip Slip), and the `finally` cleanup can then delete an attacker-selected path. Reject absolute/traversal names and resolve/normalize every entry beneath a per-import temporary root.
2. `projects.add(projectName)` plus `FileUtils.delete(new File(basePath + File.separator + projectName))` can remove an existing user directory when an archive reuses a project name or supplies a path-like name. Extraction must use a unique temporary directory and cleanup only that directory; add malicious-name and pre-existing-directory regression tests.

The current required checks are green, but these input/filesystem safety issues block merge.

tsonglew:
Thanks — all four items are addressed in `bf46f86`, on top of a rebase onto current `main`.

**1. Cache the per-project parse result.** Decryption moved to a new request-scoped `DbeaverCredentialsResolver` (`impl/ncx/dbeaver/`) that memoises the parsed document per credential file, *including* the "unavailable" outcome, so a project with 30 connections now reads, decrypts, and parses the file once instead of 30 times (and warns once, not 30 times). The service resolves it once per project, before the connection loop.

While in there: the uploaded archive and **every** extracted project are now removed in a `finally` block. Previously cleanup ran only on the success path, so a failure mid-import left extracted projects — containing decrypted credentials — on disk. The `data-sources.json` stream that was left open is closed too, which is what makes that cleanup reliable on Windows.

**2. Focused tests.** 15 new tests in `chat2db-community-domain-core`:

- `DbeaverCredentialsResolverTest` — default-key success; master-password success; missing/blank/wrong master password; missing file; malformed files (plaintext, empty, truncated, decryptable-but-not-JSON); decrypt-once-per-file caching; cached unavailable outcome; per-project isolation.
- `TaskNcxImportServiceDbeaverImportTest` — end-to-end .dbp import with the default key and with the master password; import without passwords when credentials cannot be decrypted; absent credential file; missing credential entry; malformed credential entry; multi-project cleanup on the success path **and** on the failure path.

Both new behaviours are mutation-checked: removing the cache fails `decryptsEachCredentialFileOnlyOnce`, and moving cleanup off the `finally` path fails `removesEveryExtractedProjectAndTheUploadWhenTheImportFails`.

**3. Transport/privacy decision for Community web mode.** Recorded explicitly in `SECURITY.md` under *"Credentials supplied for an import"*, and referenced from the controller javadoc. The decision: the master password stays request-scoped — used only to decrypt that archive's credential files, never written to storage, logs, or error messages, and always optional. It crosses the HTTP boundary in the same request form as the connection passwords the datasource endpoints already accept, so it **inherits** the existing loopback-only deployment boundary instead of introducing a new one; a network-reachable Community deployment remains unsupported and would require HTTPS. Notably it is handled more conservatively than existing datasource passwords, which *are* persisted — this one never is.

**4. Rebase.** Done, onto `a1f1e7b5`.

One extra find while rebasing: the placeholder string existed only in en-US and zh-CN, which fails `yarn test:i18n` — the validator requires es-ES and ko-KR key parity plus a matching `scripts/i18n-source-hashes.json`. All five locales now have it and the hash manifest is refreshed.

**Verification run:** `chat2db-community-domain-core` 186 tests / 0 failures, `chat2db-community-web` 63 tests / 0 failures (tests explicitly enabled, failure-ignore overridden), `mvn package` for both, `yarn lint` clean, `validate-i18n` clean, and `yarn build:web:community` green including the production-bundle check.

Jay17-git:
Hi Chat2DB maintainers,
I’m following up on the two security reports I submitted: Zip Slip and Zip Bomb.
Could you please provide an update on their review status and an estimated timeline for completing validation and confirming whether these issues are valid vulnerabilities?
Please let me know if you need any additional information or assistance reproducing either issue. Thank you for your time and help!

PR #2633, opened 2026-08-07
feat: universal JDBC support — add 10 databases through configuration only

Chat2DB-Pro:
Rebased replacement for #2108, which was 301 commits behind `main` and no longer merged cleanly.

## What this adds

Ten databases, all through configuration in `generic.json` — no per-database Java plugin:

| Database | Connection |
|---|---|
| Firebird | `jdbc:firebird://` |
| HSQLDB | `jdbc:hsqldb:file:` |
| Apache Derby | `jdbc:derby:` |
| QuestDB | PG wire `:8812` |
| CrateDB | PG wire `:5432` |
| YugabyteDB | PG wire `:5433` |
| Greenplum | PG wire `:5432` |
| Trino | `jdbc:trino://` |
| TimescaleDB | `jdbc:postgresql://` |
| Apache IoTDB | `jdbc:iotdb://` |

## Supporting mechanism

- `SqlSyntaxPluginRegistry` — syntax plugins resolved through a `ServiceLoader` registry
- `SupportedDatabaseRegistry` + `DbSupportedDatabaseController` — the backend serves the supported-database inventory, so the client renders newly registered databases without a rebuild
- `ConfigurableSQLIdentifierProcessor` — identifier quoting driven by an `identifierQuotes` spec (single char, `open:close` pair, or `none` for path-based dialects like IoTDB)
- Per-database vector icons, host/port auto-fill, and built-in form parity on the client
- `script/test/start-test-databases.sh` to bring up the integration-test matrix, plus integration tests for Firebird, the PG-wire family, the time-series engines, Trino, and the embedded engines

## Rebase note

`main` landed #2199 (`fix(generic): escape SQL identifiers and literals in metadata/DDL paths`) on the same `getSQLIdentifierProcessor()` override this branch replaces. Git merged both textually, leaving a duplicate method that did not compile. The final commit reconciles them: a dialect declaring `identifierQuotes` uses the configured processor, everything else falls back to `GenericIdentifierProcessor.INSTANCE` — so main's conditional ANSI quoting and string-literal escaping are preserved rather than dropped.

## Verification

- `mvn -pl chat2db-community-spi,chat2db-community-plugins/chat2db-community-generic -am compile` — passes
- Unit tests for both modules pass, including main's `GenericIdentifierProcessorTest` and this branch's `ConfigurableSQLIdentifierProcessorTest` together
- Container-backed integration tests were not run here; they need the database matrix from `script/test/start-test-databases.sh`

Closes #2108

https://claude.ai/code/session_01Rq65mLom8mVWv1F6VzejN4


Chat2DB-Pro:
Fixed in 02fc998fd — added `icon?: string | null` to `ISupportedDatabaseSummary`, matching the backend `SupportedDatabaseSummary` DTO.

Confirmed the finding was real and reproduced it with `tsc --noEmit` over the affected files: 4 errors before the fix (TS2339 at `dynamicDatabaseRegistry.ts:36`, TS2353 at test lines 43/49/106), 0 after. The reason CI stayed green is that the client has no typecheck script — `umi build` transpiles without type checking and the eslint run is not type-aware, so no job would have caught it.

PR #2636, opened 2026-08-07
feat(mysql): active InnoDB transaction inspection (#2595)

### LINKED ISSUE #2595: [MYSQL-OPS-002] Add a MySQL active transaction view
## What needs to be done

Add an active InnoDB transaction view showing transaction ID, state, start time, age, isolation level, waited lock, locked and modified row counts, thread ID, user, host, database, and current SQL. Support filtering, sorting, refresh, and navigation to the related Session.

## Acceptance criteria

- [ ] An open transaction appears after it starts, and its age continues to update.
- [ ] Waiting transactions are linked to their lock and session, and committed or rolled-back transactions disappear after refresh.
- [ ] Hidden SQL or insufficient permissions produce an explicit unavailable state rather than a misleading blank value.
- [ ] A server with no active InnoDB transaction shows a normal empty state.

Support MySQL 5.7 and 8.0 using InnoDB transaction metadata. Non-transactional engine activity is outside this view. The feature is read-only; viewing complete information for other users generally requires `PROCESS`, and SQL text must follow the existing sensitive-data display policy. Navigation to termination actions uses MYSQL-OPS-001, while this transaction list must work independently.

## Tests and test data

Add automated tests for read-write and read-only transactions, long-running transactions, lock waits, commit and rollback removal, restricted metadata, the empty state, and both MySQL 5.7 and 8.0. Provide repeatable `init.sql`, `grants.sql`, `cleanup.sql`, and `README.md` files that create the required tables, users, sessions, and transactions and document verification steps.

Aias00:
Closes #2595

## Summary
- Read-only Active Transactions view for the datasource node (right-click menu).
- Lists `information_schema.innodb_trx` joined with `processlist`: trx ID, state, start time, age, isolation level, rows locked/modified, lock structs, thread ID, user, host, database, and current SQL.
- MySQL 5.7 and 8.0 compatible; `trx_query` is null without PROCESS and rendered as an explicit unavailable state.
- Refreshable table dialog, i18n in en/es/ja/ko/zh, MYSQL-OPS-002 fixtures (PROCESS admin vs limited user).

## Test plan
- [ ] Run MYSQL-OPS-002 README verification as `ops002_admin` (open transactions, lock wait, commit removal, empty state)
- [ ] Verify limited user (`ops002_user`) sees null SQL text with the unavailable hint


openai0229:

P1: active-transaction listing does not carry the requested data-source context, so it depends on ambient connection state. Permission failures are not represented as an explicit degraded/error result. Please use a data-source-bound request, preserve transaction/session identity, add least-privilege and multi-data-source tests.

PR #2648, opened 2026-08-07
feat: add 12 more databases through configuration only

Chat2DB-Pro:
Third batch on the config-only path, following #2633 and #2644. No new Java plugin code — all entries live in `generic.json`.

## Bring their own driver

| Database | Driver class | Port |
|---|---|---|
| SAP HANA | `com.sap.db.jdbc.Driver` | 30015 |
| MonetDB | `nl.cwi.monetdb.jdbc.MonetDriver` | 50000 |
| Altibase | `Altibase.jdbc.driver.AltibaseDriver` | 20300 |
| Ocient | `com.ocient.jdbc.JDBCDriver` | 4050 |
| InterSystems IRIS | `com.intersystems.jdbc.IRISDriver` | 1972 |
| NuoDB | `com.nuodb.jdbc.Driver` | — |
| OpenSearch | `org.opensearch.jdbc.Driver` | 9200 |
| DolphinDB | `com.dolphindb.jdbc.Driver` | 8848 |
| Db2 for IBM i | `com.ibm.as400.access.AS400JDBCDriver` | — |

## Reuse a jar we already bundle

| Database | Reuses | Why | Port |
|---|---|---|---|
| GreptimeDB | MySQL driver | speaks the MySQL wire protocol | 4002 |
| Apache Spark | Hive driver | speaks HiveServer2 | 10000 |
| TimechoDB | IoTDB driver | is an IoTDB distribution | 6667 |

Two entries get non-default handling: OpenSearch is routed to the Elasticsearch syntax plugin, and TimechoDB inherits the IoTDB entry's quoting rules (`identifierQuotes: none`, table-only qualification) because dots are structural in its path-based references and quoting any part would break the SQL.

## How the values were established

Same method as the previous two batches — nothing written from memory:

- **Driver classes** — every jar was downloaded from Maven Central and its `META-INF/services/java.sql.Driver` read directly. MonetDB ships no service file, so its class was confirmed by listing the jar contents; the class matches what DBeaver declares.
- **Coordinates** — found by reverse-searching Maven Central on the fully-qualified driver class (`fc:` queries) rather than guessing artifact names.
- **Ports** — from DBeaver's own `plugin.xml` driver metadata (`defaultPort`).

## Deliberate omissions

- **Apache Kyuubi** — the published `kyuubi-hive-jdbc` jar is unshaded and cannot load standalone. Shipping it would hand users a driver that fails at connect time.
- **Netezza, SQream, Timeplus** — no driver on Maven Central (zero results).
- **Athena, Cloudera Impala** — vendor-distributed drivers, not on Maven Central.

These belong to the "user supplies their own jar" path rather than the config-only path.

## Progress against DBeaver

A gap analysis against DBeaver's repository (183 driver declarations across 75 ext plugins → 125 distinct products) put us at 44 supported. After #2644 and this PR we are at **69, with the gap down from 82 to 57**.

What remains is mostly not reachable by configuration alone: ~23 commercially licensed drivers, ~11 non-JDBC engines needing dedicated plugins, and a set of cloud services requiring SDKs and credentials. The cheap wins from this mechanism are largely spent; the highest-leverage next step is a "bring your own driver jar" flow, which would unlock the licensed group in one change.

## Verification

- `generic.json` grows from 26 to 38 entries; additive diff only, no reformatting
- generic plugin unit tests pass, including `GenericPluginConfigLoadTest` which parses the file
- Live connections were not exercised; these are configuration entries and need the respective engines to test end to end

https://claude.ai/code/session_01Rq65mLom8mVWv1F6VzejN4


Chat2DB-Pro:
Fixed in `1a4c4bd15`, but the finding needs correcting on the reason.

**The stated reason does not hold.** I ran the driver's own `UrlParser.parseProperties` against the shipped template:

```
jdbc:opensearch://localhost:9200         -> {port=9200, host=localhost, usessl=false}
jdbc:opensearch://https://localhost:9200 -> {port=9200, host=localhost, usessl=true}
jdbc:opensearch://http://localhost:9200  -> {port=9200, host=localhost, usessl=false}
```

The scheme is optional and defaults to plain HTTP — which is also why DBeaver brackets it as `[{scheme}://]` in its own sample URL. So the URL was not the problem.

**The conclusion was right, for a different reason.** Loading the driver and calling `getConnection` shows it fails *before reaching the network*:

```
NoClassDefFoundError: com/amazonaws/auth/AWSCredentialsProvider
```

The published jar is unshaded. That is a real defect and the entry had to go.

**The check was worth generalising.** I ran the same load-and-connect probe across all 12 entries in this PR, which found two more the review did not:

| Entry | Result |
|---|---|
| OpenSearch | fails on AWS SDK |
| Ocient | fails on `log4j-core` |
| DolphinDB | fails on `com.xxdb.data.Vector` |
| HANA, MonetDB, Altibase, IRIS, NuoDB, Db2 for IBM i | reach the network |
| GreptimeDB, Apache Spark, TimechoDB | reuse already-bundled drivers |

My original verification only confirmed each jar declares a `java.sql.Driver` service, which does not prove the driver can load. That is exactly the defect Apache Kyuubi was excluded for in this same PR, so I applied the same standard.

**Resolution:**

- **DolphinDB kept** — its closure is just two jars (`dolphindb-javaapi`, `slf4j-api`), added via the existing multi-jar mechanism (same shape as the Derby entry). Re-probed: reaches the network.
- **OpenSearch and Ocient removed** — AWS SDK and netty/commons-text closures respectively, not worth pinning by hand.

I also probed the drivers in #2649 and the already-merged #2644; all of those reach the network, so no further entries are affected.

PR #2676, opened 2026-08-11
fix(snowflake): keep connection properties idempotent

### LINKED ISSUE #2675: bug(snowflake): connection properties accumulate or fail on reconnect
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

be8be02b59eb7ae16fd2af3717225ef428c58edb (source build)

### Deployment

Web

### Operating System

Windows

### Operating System Version

Windows 11 10.0.26200

### Database and Version

Snowflake JDBC; the defect is in local connection-property assembly and was reproduced with a capturing JDBC driver without a live service.

### Problem Summary

`SnowflakeDBManager.getConnection` mutates `ConnectInfo.extendInfo` in place. A null or immutable list fails before connection, while reconnecting with a mutable list repeatedly appends Snowflake-managed properties and can also hide DriverConfig fallback properties.

### Steps to Reproduce

1. Create a Snowflake `ConnectInfo` with database/schema values and `extendInfo` set to null, `List.of()`, or a mutable property list.
2. Configure DriverConfig fallback properties such as `warehouse` and `role`.
3. Call `SnowflakeDBManager.getConnection` using a capturing JDBC driver.
4. For the mutable case, call it again with the same `ConnectInfo` and inspect `extendInfo`.

### Expected Behavior

Connection properties are assembled from a mutable copy, DriverConfig fallback values are preserved, and `db`, `schema`, and `JDBC_QUERY_RESULT_FORMAT` each occur once with current values on every reconnect.

### Actual Behavior

Null input throws `NullPointerException`, immutable input throws `UnsupportedOperationException`, and mutable input accumulates duplicate managed properties on reconnect. Adding the first managed property also prevents the normal DriverConfig fallback from being used.

### Logs

```shell
Focused regression: SnowflakeDBManagerTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
```

Fork validation PR: HandSonic/Chat2DB#28. The tests use a capturing JDBC driver and verify the exact properties passed to a new connection as well as reconnect idempotency.

### Impact

Degrades a core workflow. Connection creation can fail for valid configuration shapes, and repeated reconnects can send stale or duplicate JDBC properties.

### Workaround

A partial workaround exists: callers can prebuild a mutable explicit property list and avoid reusing `ConnectInfo` across reconnects.

## Submission Checklist

HandSonic:
## Related issue

Fixes #2675

## Summary

- Copy connection properties before adding Snowflake-managed values, so null, empty, and immutable lists are safe.
- Replace `db`, `schema`, and `JDBC_QUERY_RESULT_FORMAT` case-insensitively instead of appending duplicates on reconnect.
- Remove stale `db` and `schema` properties when the current database or schema is cleared.
- Preserve DriverConfig fallback properties when explicit `extendInfo` is null or empty.

## Affected surfaces

- [ ] Frontend / Web
- [x] Backend / API / Storage
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Latest `main` baseline reproduction: 2 tests failed as expected. Reconnecting duplicated `db`, and immutable properties threw `UnsupportedOperationException`.
- `SnowflakeDBManagerTest`: 3 tests passed, covering immutable and duplicate properties, idempotent reconnects, cleared database/schema values, and null/empty DriverConfig fallback.
- Snowflake plugin full suite on latest `main`: 38 tests passed with zero failures, errors, or skips.
- Dependency reactor: 827 tests passed with zero failures or errors.
- `git diff --check`: passed.

## Risk and compatibility

No API or persisted-data format changes. The change only normalizes the connection properties passed to the Snowflake JDBC driver. A real Snowflake server is not required for this deterministic pre-connection transformation.

## Reviewer map

- Start with `SnowflakeDBManager.prepareExtendInfo` and `replaceProperty`.
- Verify that non-managed properties retain order, managed keys are unique and canonical, blank database/schema values remove stale keys, and input lists are not mutated.
- Rollback requires no migration.

## Contributor declaration

- [x] I linked the issue that defines this change.
- [x] I disclosed substantial AI assistance below.

AI assistance: OpenAI Codex assisted with baseline reproduction, implementation reduction, focused tests, and review.


openai0229 (dismissed):
Blocking idempotency case:  only removes/replaces a managed property when the new value is nonblank. After the first call stores  and  back into the reused , clearing  or  and reconnecting leaves the previous property in place. The connection can therefore retain a stale database/schema context.\n\nPlease always remove the managed / entries first, then add the property only when the current value is nonblank. Add regression coverage for reusing one  while changing each value from nonblank to blank. The mutable-copy and DriverConfig fallback handling otherwise look correct.

openai0229 (changes requested):
Blocking idempotency case: `replacePropertyIfConfigured` only removes/replaces a managed property when the new value is nonblank. After the first call stores `db` and `schema` back into the reused `ConnectInfo.extendInfo`, clearing `databaseName` or `schemaName` and reconnecting leaves the previous property in place. The connection can therefore retain a stale database/schema context.

Please always remove the managed `db`/`schema` entries first, then add the property only when the current value is nonblank. Add regression coverage for reusing one `ConnectInfo` while changing each value from nonblank to blank. The mutable-copy and DriverConfig fallback handling otherwise look correct.

HandSonic:
Addressed on the current head: managed db/schema entries are always removed first and only re-added for nonblank current values. Reusing one ConnectInfo while clearing either or both values is covered. The full Snowflake suite passes (40 tests) on current main. Please re-review when convenient.

openai0229 (approved):
Re-reviewed optimized head 1666945a9e3f1b328b90cb4cb8dd042fefb63a39. The production path now uses one case-insensitive replace-or-remove helper, and focused tests cover immutable input, duplicate managed keys, idempotent reuse, cleared database/schema values, and DriverConfig fallback without mutating global JDBC state. Baseline reproduction failed as expected; focused and full Snowflake suites passed on latest main.

PR #2687, opened 2026-08-12
fix(tools): harden asynchronous JDBC downloads

### LINKED ISSUE #2684: bug(jdbc): async driver downloads keep the JVM alive and expose partial jars
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

main at 2442df98d92768bbdaf4ac9a120938f4014d5d53

### Deployment

Web, Desktop, and embedded backend

### Operating System

Any

### Problem Summary

JdbcJarUtils configures the asynchronous OkHttp Dispatcher with Executors.newFixedThreadPool(20). The default thread factory creates non-daemon workers and the static executor is never shut down, so the first asynchronous driver download can keep the JVM alive after the application context closes. Downloads also write directly to the final jar path, which makes an interrupted process expose a partial driver file.

### Steps to Reproduce

1. Remove a startup JDBC driver jar so JarDownloadTask triggers JdbcJarUtils.asyncDownload.
2. Let the asynchronous request start or complete.
3. Close the Spring or embedded application context without calling System.exit.
4. Observe the JVM process and the driver directory during an interrupted download.

### Expected Behavior

Download workers do not prevent normal JVM shutdown, and only complete driver jars are published at their final paths.

### Actual Behavior

Up to 20 non-daemon fixed-pool workers remain alive indefinitely. A forced shutdown can also leave a truncated file at the final jar path, and the next startup may treat its existence as a completed download.

### Logs

No error is required; the worker threads wait on the executor queue after the request completes.

The executor is static, core threads do not time out, and no shutdown path exists. Any daemon-thread fix must publish through a temporary file before atomically replacing the final jar to avoid making interrupted downloads visible.

### Impact

Natural shutdown can hang in tests, embedded use, or controlled application exits. Interrupted downloads can leave an unusable JDBC driver cached across restarts.

### Workaround

Force-terminate the JVM and manually delete any truncated driver jar before restarting.

### Submission Checklist

- [x] Reproduced on the selected Chat2DB edition.
- [x] Searched existing issues and pull requests for duplicates.
- [x] Removed credentials and private data.

HandSonic:
## Summary

- Use daemon workers for asynchronous JDBC driver downloads so they cannot keep the JVM alive.
- Publish completed downloads atomically and isolate concurrent temporary-file cleanup.

## Validation

- Focused tests: 9/9
- Tools module tests: 72/72
- Maven package

Fixes #2684

## Latest-main verification (2026-09-04)
- Rebased onto `144a04ee2`; full tools suite passed (73 tests).`n- The Linux CI-only race was traced to OkHttp temporarily renaming a daemon worker during its callback. The test now separates stable factory naming from actual callback daemon status; the focused suite passed 10 consecutive runs before the full suite.`n- Combined with #2695 using a semantic conflict resolution; full combined tools suite passed (76 tests), covering daemon workers, atomic publication, and dynamic runtime paths.

### DIFF
--- a/chat2db-community-server/chat2db-community-tools/src/main/java/ai/chat2db/community/tools/util/JdbcJarUtils.java
+++ b/chat2db-community-server/chat2db-community-tools/src/main/java/ai/chat2db/community/tools/util/JdbcJarUtils.java
@@ -9,9 +9,15 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.List;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 
 import cn.hutool.core.io.FileUtil;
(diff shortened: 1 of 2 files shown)


openai0229 (dismissed):
The daemon dispatcher and unique temporary-file plus atomic publish flow fix both reported failure modes, including cleanup after interrupted writes. Approved.\n\nNon-blocking follow-up: the response stream still has no maximum byte limit, so a future hardening change should cap driver downloads and delete the temporary file on overflow. Merge this before #2695, then rebase #2695 so its configured-URL changes retain this atomic-write behavior.

openai0229 (dismissed):
The daemon dispatcher and unique temporary-file plus atomic publish flow fix both reported failure modes, including cleanup after interrupted writes. Approved.

Non-blocking follow-up: the response stream still has no maximum byte limit, so a future hardening change should cap driver downloads and delete the temporary file on overflow. Merge this before #2695, then rebase #2695 so its configured-URL changes retain this atomic-write behavior.

openai0229 (approved):

PR #2689, opened 2026-08-12
fix(redis): persist no-expiration TTL updates

### LINKED ISSUE #2681: bug(redis): removing a key TTL does not persist the key
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

main at 2442df98d92768bbdaf4ac9a120938f4014d5d53

### Deployment

Web and Desktop

### Problem Summary

The Redis editor uses TTL -1 to mean no expiration. Updating an existing key from a positive TTL to -1 does not emit Redis PERSIST, so the original TTL keeps counting down and the key still expires.

### Steps to Reproduce

1. Create a Redis string key with TTL 60.
2. Edit the key without changing its name or value.
3. Change TTL to -1 and save.
4. Query TTL for the key.

### Expected Behavior

The update sends PERSIST for the edited key and TTL returns -1.

### Actual Behavior

The update emits no TTL command, so the previous positive TTL remains active.

### Logs

No exception is raised; the update silently leaves the old TTL in place.

RedisScriptExecutor.update only emits EXPIRE when the new TTL is greater than zero. The no-expiration transition is therefore skipped. This is directly reproducible from the current update control flow and command-generation tests.

### Impact

A key that the user explicitly marks as persistent can still be deleted when its old TTL expires.

### Workaround

Run PERSIST manually after editing the key.

### Submission Checklist

- [x] Reproduced on the selected Chat2DB edition.
- [x] Searched existing issues and pull requests for duplicates.
- [x] Removed credentials and private data.

HandSonic:
## Summary

- Send Redis PERSIST when an existing key changes from a positive TTL to no expiration.
- Add regression coverage for TTL command selection.

## Tests

- RedisScriptExecutorUpdateTest: 7/7
- Redis plugin test suite: 58/58
- Redis plugin package build

Fixes #2681


openai0229 (dismissed):
The TTL transition logic is correct: a positive TTL still emits , and changing an existing non-permanent key to  emits  without affecting type-replacement flows. Tests cover the command selection. Approved.

openai0229 (dismissed):
The TTL transition logic is correct: a positive TTL still emits `EXPIRE`, and changing an existing non-permanent key to `-1` emits `PERSIST` without affecting type-replacement flows. Tests cover the command selection. Approved.

openai0229 (approved):
已基于最新主线复核并完成 Redis 7 实测。TTL 更新现在由用户提交的新值决定：正数执行 EXPIRE，-1 执行 PERSIST；不再依赖可能过期的旧 TTL 快照。聚焦测试 8/8、Redis 插件测试 60/60、Community 后端 48 模块打包及 Playwright 正数、转 -1、陈旧快照、取消路径均通过。

PR #2690, opened 2026-08-12
fix(clickhouse): preserve IPv6 URL when switching schema

HandSonic:
## Summary

- Preserve the host and port of IPv6 ClickHouse JDBC URLs when switching schemas.
- Keep query and fragment components intact during the rewrite.

## Validation

- Focused regression tests: 5/5
- ClickHouse module tests: 48/48
- Maven package

Fixes #2685

Issue #2685, opened 2026-08-11
bug(clickhouse): switching schema corrupts IPv6 JDBC URLs


HandSonic:
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

main at 2442df98d92768bbdaf4ac9a120938f4014d5d53

### Deployment

Web and Desktop

### Operating System

Any

### Database and Version

ClickHouse with JDBC access over IPv6

### Problem Summary

Switching a ClickHouse connection to another schema corrupts bracketed IPv6 JDBC URLs. ClickHouseDBManager splits the URL using the bare host and port as a regular expression, which cannot match the bracketed IPv6 authority and does not safely preserve URL query or fragment suffixes.

### Steps to Reproduce

1. Configure a ClickHouse connection with URL jdbc:clickhouse://[2001:db8::1]:8123/default?ssl=true.
2. Store the connection host as 2001:db8::1 and port as 8123.
3. Switch the connection schema to analytics.

### Expected Behavior

The URL becomes jdbc:clickhouse://[2001:db8::1]:8123/analytics?ssl=true and remains stable on reconnect. Query parameters and ClickHouse fragment tags are preserved.

### Actual Behavior

The host and port split does not match the bracketed authority. The implementation appends the bare address and schema to the existing URL, producing a malformed value such as jdbc:clickhouse://[2001:db8::1]:8123/default?ssl=true2001:db8::1:8123/analytics. The bad URL is then written back to ConnectInfo.

### Logs

The ClickHouse driver rejects the malformed JDBC URL or connects with the wrong database.

### Screenshots or Additional Context

ClickHouse supports schema switching and fragment tags. URL rewriting should replace only the database path while preserving the authority, query, and fragment boundaries.

### Impact

IPv6 ClickHouse connections cannot reliably switch databases and reconnect attempts can reuse the corrupted URL.

### Workaround

Manually restore the JDBC URL and avoid switching schemas from Chat2DB.

### Submission Checklist

- [x] Reproduced on the selected Chat2DB edition.
- [x] Searched existing issues and pull requests for duplicates.
- [x] Removed credentials and private data.

PR #2692, opened 2026-08-12
fix(clickhouse): encode schema names in JDBC URLs

HandSonic:
## Summary

- Encode the selected ClickHouse schema as one UTF-8 JDBC URL path segment before reconnecting.
- Prevent /, ?, #, spaces, Unicode, plus signs, and literal percent text from changing URL structure or the selected schema.
- Preserve the existing JDBC query suffix and add focused regression coverage.

## Validation

- Focused regression tests: 3 passed.
- Full ClickHouse module tests: 46 passed, 0 failed.
- ClickHouse reactor package: passed.
- Fork code CI: frontend, backend, repository checks, Java/JavaScript CodeQL, SBOM, and license checks passed.
- Independent adversarial review: no findings.

Follow-up regression fix related to #2685 for schema names containing reserved URL characters. This intentionally does not close that issue; the separate IPv6 URL fix is in #2690.

## Latest-main verification (2026-09-04)
- Rebased onto `144a04ee2`; full ClickHouse plugin suite passed (55 tests), including IPv6, user-info, query/fragment, schema encoding, and idempotence cases.`n- `git diff --check` and pairwise integration analysis passed.


openai0229 (dismissed):
Blocking integration issue: this branch still applies encoding on top of the old  implementation. That implementation is exactly what #2690 replaces because it cannot reliably preserve IPv6 authorities, no-path URLs, fragments, and other URL shapes. Merging the two fixes independently leaves overlapping implementations and an avoidable conflict.\n\nPlease base this change on #2690's  helper and encode the schema as one UTF-8 path segment inside that helper. Keep the reserved-character tests from this PR and the IPv6/query/fragment tests from #2690 so there is one URL-rewrite implementation.

openai0229 (changes requested):
Blocking integration issue: this branch still applies encoding on top of the old `url.split(connectAddress)` implementation. That implementation is exactly what #2690 replaces because it cannot reliably preserve IPv6 authorities, no-path URLs, fragments, and other URL shapes. Merging the two fixes independently leaves overlapping implementations and an avoidable conflict.

Please base this change on #2690's `replaceDatabaseInJdbcUrl` helper and encode the schema as one UTF-8 path segment inside that helper. Keep the reserved-character tests from this PR and the IPv6/query/fragment tests from #2690 so there is one URL-rewrite implementation.

HandSonic:
Addressed on the current head: there is now one 
eplaceDatabaseInJdbcUrl implementation with both the schema-as-one-UTF-8-segment cases and the IPv6/no-path/query/fragment/idempotence cases. The strict-subset #2690 was closed. The full ClickHouse suite passes (55 tests) on current main. Please re-review when convenient.

openai0229 (approved):

PR #2693, opened 2026-08-12
feat(desktop): expose pluggable update controls

openai0229:
## Summary
- expose a pluggable desktop updater contract for Enterprise
- add Beta preference and update recovery UI
- preserve Community legacy updater behavior

## Verification
- Community JCEF tests: 49/49
- Community client tools: 67/67
- hot-update test and i18n checks passed


openai0229:
Fixed in 28ffc779: accepted installation now continues into the existing restart flow; rejected and failed installations remain in the failed state. Added regression assertions for both paths, and `yarn test:hot-update` passes.

openai0229:
The rerun backend failure is unrelated to this PR: `ConfigOnlyDatabaseApiFlowTest` depends on `https://cdn.chat2db-ai.com/lib/hsqldb-2.7.3.jar`, which currently returns HTTP 404. The same PR passed the complete backend job before the frontend-only regression-test follow-up; all updater-specific local suites pass.

PR #2707, opened 2026-08-18
fix(database): use server connection for MySQL deletion

openai0229:
## Related issue

N/A - regression identified while validating MySQL database deletion in the 5.3.4 code path.

## Summary

MySQL database deletion currently receives a connection context whose active database is also the delete target. The existing safety guard then rejects every normal delete request before `dropDatabase` runs.

This change restores the intended non-target connection contract:

- rebind MySQL database prepare and execute operations to a server-level connection with no selected database;
- retain the existing PostgreSQL `postgres` maintenance-database behavior;
- keep the current-target safety guard in place;
- document that `rebindCurrentDatabase(null)` means a server-level connection;
- add a regression test that begins with the connection profile targeting the database being deleted.

## Affected surfaces

- [ ] Frontend / Web
- [x] Backend / API / Storage
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Commands and results:
  - `mvn -pl chat2db-community-domain/chat2db-community-domain-core -am -Dmaven.test.skip=false -Dtest=DatabaseObjectDeleteServiceImplTest -Dsurefire.failIfNoSpecifiedTests=false test` - 13 tests passed, 0 failures, reactor build successful.
  - `mvn -pl chat2db-community-domain/chat2db-community-domain-core -am -Dmaven.test.skip=false test` - 123 domain-core tests passed, 0 failures, reactor build successful.
  - `git diff --check` - passed.
- Manual verification: Source-path verification only; no live MySQL database was deleted.
- UI evidence: N/A

## Risk and compatibility

- Public API or stored data: No public endpoint or stored-data contract changes. The internal connection-context Javadoc now documents the existing nullable server-level scope.
- Database or driver compatibility: MySQL database deletion now opens a server-level connection. PostgreSQL keeps using the `postgres` maintenance database; schema deletion is unchanged.
- Network, privacy, or security: No new network destinations, credentials, or data exposure. The existing current-target safety guard remains enabled.
- Community / Local / Pro boundary: Community backend change; downstream editions that consume this backend inherit the corrected deletion behavior.
- Backward compatibility: Restores the pre-migration connection semantics for MySQL database deletion without changing request or response payloads.

## Reviewer map

- Start here: `DbDatabaseObjectDeleteServiceImpl.reconnectForDatabaseDelete`, then `mySqlDatabaseDeleteRebindsFromTargetDatabaseToServerLevelConnection`.
- Failure condition: A MySQL delete request whose initial profile targets the delete database still throws `database.delete.connectionTargetsDeletedDatabase`, or reaches the driver with that database still selected.
- Rollback or disable path: Revert this commit. The safety guard remains independent of the rebind change.

## Contributor declaration

AI assistance: Substantial AI assistance was used for root-cause analysis, implementation, regression-test design, and PR drafting. The resulting changes were reviewed through the focused and full reactor test commands listed above.


openai0229 (commented):
Review completed after rebasing onto the latest main.

No blocking issue found. The MySQL prepare and execute paths now consistently reconnect at server scope, while PostgreSQL keeps its maintenance-database behavior. The production DbConnectionContextServiceImpl also supports a null database name, so the deletion flow is coherent end to end.

Non-blocking test suggestion: add a prepare-from-target-database case so both entry contexts are covered explicitly.

openai0229:
Focused verification completed after review: DatabaseObjectDeleteServiceImplTest passed 13/13 with the domain-core reactor build successful.

PR #2739, opened 2026-08-23
fix(export): isolate skipped tables and avoid colliding ZIP entries

HandSonic:
## Related issue

N/A — export regressions identified while reviewing the task-system rebuild in #2712.

## Summary

Database SQL exports through `DefaultDBManager` now skip unreadable tables without retaining their partial SQL. Each table is staged on disk and appended to the task artifact only after its export completes. A read failure after rows have been exported discards that table's DROP/CREATE/INSERT statements and records an `OBJECT_SKIPPED` warning. Cancellation and file I/O failures still abort the task; single-table export failures still propagate.

Multi-table data exports now choose temporary filenames against the actual filesystem. Tables such as `db1/users`, `db2/users`, and `USERS-2` produce separate ZIP entries with their original data, including on case-insensitive filesystems.

## Affected surfaces

- [ ] Frontend / Web
- [x] Backend / API / Storage
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Before the fixes, two added regression tests failed: a skipped table retained destructive partial SQL, and a generated filename collided with an existing case alias and overwrote ZIP content. Both pass after the fixes.
- Full affected modules and dependencies: `mvn -o -B -f chat2db-community-server/pom.xml -pl :chat2db-community-spi,:chat2db-community-domain-core -am -Dmaven.test.skip=false -DskipTests=false '-Dsurefire.includes=**/*Test.java' -Dmaven.test.failure.ignore=false test` — SPI 149/149 and domain-core 250/250 passed; dependency tests passed.
- Backend package: `mvn -o -B -f chat2db-community-server/pom.xml -pl chat2db-community-start -am -Dmaven.test.skip=true -Dchat2db.finalName=chat2db-community package` — passed; this package command skips tests.
- Live JDBC harness: exercised the generic manager with PostgreSQL data and a query that fails after the first 100,000 exported rows. The failed table was absent from the final SQL, and the following table was complete.
- Playwright CLI: opened the Community Web export dialog, cancelled without submitting, then completed a single-table CSV export. Submitted a multi-table request to the same `/api/tasks/export` endpoint and verified all three ZIP entries and their distinct row contents. The current UI exposes single-table data export; the multi-table check used the API.
- Regression coverage also includes cancellation, final-artifact write failure, wrapped staging I/O failure, and large Unicode/multiline values.
- UI evidence: runtime tests completed on separate loopback ports. Temporary browser artifacts and test data are excluded from Git. Existing React/rc-virtual-list development warnings were observed.

## Risk and compatibility

- Public API or stored data: no interface or persisted-format changes.
- Database or driver compatibility: the per-table skip behavior applies to the generic database-export loop. Database-specific overrides retain their current behavior.
- Disk use: one table is staged at a time, requiring temporary disk space for that table. Normal completion, failure, and cancellation close and delete its temporary file.
- Network, privacy, or security: no added network calls. The staging file is created by the platform temporary-file API.
- Community / Local / Pro boundary: N/A.
- Backward compatibility: single-table failures remain failures; successful SQL write boundaries and content are preserved. ZIP filenames receive a suffix only when they collide on the output filesystem.

## Reviewer map

- `DefaultDBManager.exportTables` and package-private `TableExportContext`: stage, skip, publish, and abort boundaries.
- `BaseExporter.uniqueTableFile`: filesystem-aware collision detection.
- `DefaultDBManagerDatabaseExportToleranceTest` and `BaseExporterTest`: partial-output, write-failure, cancellation, and filename-collision regressions.
- Rollback: revert this PR.

## Contributor declaration

AI assistance: substantial — the initial fixes and the follow-up review fixes/tests were developed with AI assistance. Local module tests, a real JDBC failure scenario, and Community Web/API export verification were executed.


openai0229:
Thanks @HandSonic for identifying these export regressions and for your work on this PR.

After reviewing the implementation and trying the current UI, we have decided to design scope selection, output options, and failure handling together in a unified export wizard. Users should be able to choose whether to stop on an error or continue, and clearly see which objects completed, failed, or were skipped.

We opened #2894 to track this work, and the maintainers will lead the design and implementation. We are closing this PR in favor of that broader work to avoid further rework for you. No additional changes are requested from you on this PR.

The filename-collision fixes, protection against partial table output, and regression cases remain valuable. We have linked this PR and its reviewed commit in #2894 so we can reuse the applicable work and preserve attribution when integrating the new workflow.

PR #2797, opened 2026-08-29
fix(sql): preserve result context during paging

HandSonic:
Runs pagination, filtering, and refresh against the clicked result data source, replaces that result in place without starting a new execution batch, preserves sibling and pinned history tabs, filters mixed stream events by result-set identity, surfaces failed/no-match responses, and rolls back optimistic paging on cancellation. Includes focused web, desktop, history, and cancellation coverage.


HandSonic:
Closing as superseded by current main. Result paging is now intentionally owned inside ResultSet, and the current regression suite explicitly asserts that SQLExecute must not wire the old parent onResultPagingChange callback. Rebasing this PR leaves that legacy handler unused and fails ESLint; reconnecting it would restore the obsolete architecture.

PR #2913, opened 2026-09-15
fix(redis): support RedisJSON keys

### LINKED ISSUE #2674: 不支持Redis的json数据
### Chat2DB Edition

Chat2DB Community

### Chat2DB Version

5.3.3

### Deployment

Desktop

### Operating System

macOS

### Operating System Version

MAC

### Database and Version

_No response_

### Problem Summary

打不开redis的json数据，报错

### Steps to Reproduce

打开redis的json数据报错

### Expected Behavior

能打开json数据

### Actual Behavior

ok

### Logs

```shell
""
```

Image

### Impact

Blocks startup or a core workflow

### Workaround

Not sure

### Submission Checklist

lorenzozanee:
## Related issue

Closes #2674

## Summary

RedisJSON module keys were treated as unknown Redis strings and opened with `GET`. This change recognizes RedisJSON type aliases, uses `JSON.GET`/`JSON.SET` with separate expiry handling, and keeps ordinary Redis strings on `GET`.

## Affected surfaces

- [ ] Frontend / Web
- [ ] Backend / API / Storage
- [ ] CI / Build / Release
- [ ] Documentation only

## Verification

- Commands and results: `git diff --check` passed. Focused Maven tests and frontend tests/lint were not run because Java/Maven and the required frontend executables are unavailable.
- Manual verification: Not run; no RedisJSON service is available.
- UI evidence: N/A

## Risk and compatibility

- Public API or stored data: The API exposes the normalized `json` type; no stored-data migration.
- Database or driver compatibility: RedisJSON keys require the RedisJSON command set; ordinary Redis core types are unchanged.
- Network, privacy, or security: N/A.
- Community / Local / Pro boundary: Community Redis plugin and Redis browser only.
- Backward compatibility: Existing Redis types retain their existing commands and payload shapes.

## Reviewer map

- Start here: Redis type normalization, JSON command handling, and Redis detail dispatch.
- Failure condition: RedisJSON keys were treated as unknown strings and opened with `GET`.
- Rollback or disable path: Revert the RedisJSON type and editor changes; existing core Redis type handling is independent.

## Contributor declaration

- [ ] I tested the affected behavior and reported the actual results above.
- [ ] I did not include credentials, private data, or generated build output.

Supersedes #2897
原PR分支已删，按原提交重建


openai0229 (commented):
Please address the three inline findings before merging: preserve valid JSON on reads, validate JSON before destructive type conversion, and correct the failing test expectations.

Validation on this head: I compiled and directly executed the five added Java test methods in an isolated harness using local dependencies (2 passed, 3 failed). I also exercised the JSON.GET result parser and JDBC ResultSet from the configured redis-jdbc-driver-1.5.jar. The type-conversion data-loss finding is based on the command execution path; it has not been reproduced against a live RedisJSON server. Full Maven, live RedisJSON, and desktop UI verification remain outstanding.

chat2db-community-server/chat2db-community-plugins/chat2db-community-redis/src/main/java/ai/chat2db/plugin/redis/type/JsonTypeScript.java
```diff
@@ -0,0 +1,83 @@
...
+    public RedisKey getKeyR(Connection connection, RedisKey redisKey) {
+        if (!existKey(connection, redisKey.getName())) {
+            return null;
+        }
+        RedisKey result = new RedisKey();
+        result.setName(redisKey.getName());
+        result.setType(RedisDataType.JSON.getCode());
+        DefaultSQLExecutor.getInstance().execute(connection, getKey(redisKey), resultSet -> {
+            while (resultSet.next()) {
+                String value = resultSet.getString(RedisConstants.FIELD_VALUE);
```
openai0229, line 33: [P1] Preserve the JSON document instead of the driver's Java string representation

The configured redis-jdbc-driver-1.5.jar decodes JSON.GET responses before exposing them through ResultSet. An isolated check of that actual driver shows that an input document `{"name":"alice"}` becomes `{name=alice}` through getString("value"), and the JSON string `"hello"` becomes `hello`. JSON null also becomes Java null and is skipped by the following condition. Consequently the editor receives invalid JSON or no value payload, and edited object/string documents cannot be written back successfully.

Please preserve the original JSON text, including numeric precision and JSON null, rather than relying on getString() for decoded objects. Add coverage for objects, nested arrays, strings, numbers, booleans, and null through the real JDBC result handling, plus a read/edit/save round trip against RedisJSON.

chat2db-community-server/chat2db-community-plugins/chat2db-community-redis/src/main/java/ai/chat2db/plugin/redis/type/JsonTypeScript.java
```diff
@@ -0,0 +1,83 @@
...
+        result.setTtl(StringUtils.isNotBlank(ttl) ? Long.parseLong(ttl) : -1L);
+        return result;
+    }
+
+    @Override
+    public List<String> createKey(RedisKey redisKey) {
+        if (redisKey == null || redisKey.getValue() == null) {
+            return List.of();
+        }
+        String script = jsonSet(redisKey.getName(), redisKey.getValue());
```
openai0229, line 49: [P1] Validate JSON before allowing an existing key to be converted

createKey() currently accepts invalid JSON such as `not-json` and returns a JSON.SET command without error. In RedisScriptExecutor.update(), changing an existing key's type first generates this command, then queues DEL for the old key before JSON.SET, and executes those commands sequentially. Selecting the newly added json type for an existing string whose value is not valid JSON therefore reaches a path that deletes the old key before Redis rejects the replacement.

Please parse and validate the JSON before generating commands for this conversion, so invalid input fails before DEL can run. The same JSON preparation should accept valid formatted JSON: currently raw CR/LF characters are rejected by getRedisValue(). Add a regression showing that invalid conversion preserves the original key, value, and TTL; keep the fix scoped to this new JSON path.

chat2db-community-server/chat2db-community-plugins/chat2db-community-redis/src/test/java/ai/chat2db/plugin/redis/type/JsonTypeScriptTest.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Test
+    void readsTheJsonDocumentWithJsonGet() {
+        assertEquals("JSON.GET 'profile'", typeScript.getKey(key("profile", null)));
+    }
+
+    @Test
+    void createsTheJsonDocumentWithJsonSet() {
+        assertEquals(List.of("JSON.SET 'profile' '$' '{\"name\":\"alice\"}'"),
```
openai0229, line 25: [P2] Correct the three JSON.SET test expectations and run them

I compiled and directly invoked all five added Java test methods in an isolated harness. readsTheJsonDocumentWithJsonGet and mapsRedisJsonModuleTypesToJsonScript passed; createsTheJsonDocumentWithJsonSet, preservesTheRequestedTtlWhenCreatingTheJsonDocument, and updatesTheJsonDocumentWithJsonSet failed.

The expected commands here and at lines 34 and 41 omit the backslashes that the existing getRedisValue() adds before double quotes. Please correct the expected Java string literals while preserving the JDBC argument escaping, then run the focused Maven tests with test execution enabled and failure-ignore disabled. Command-string assertions should be supplemented by the real JSON read/write coverage described above.

PR #2933, opened 2026-09-18
fix(community): keep the bundled update key resource product specific

openai0229:
## Why

The previous unification renamed Community's bundled update signing key to `chat2db-update-keys.properties` — the same resource name the Studio desktop ships inside `chat2db-studio-jcef`. A Studio desktop carries **both** jars on `runtime/lib` (`chat2db-community-jcef-5.3.0.jar` and `chat2db-studio-jcef-5.3.0.jar`, verified in the Studio build) and `TrustedUpdateKeys.load()` resolves the resource through the classpath, so the jar that comes first wins — and `chat2db-community-*` sorts first.

Probe with a tiny program on a classpath holding both jars:

```
java -cp chat2db-community-jcef.jar:chat2db-studio-jcef.jar Probe
-> url=jar:...chat2db-community-jcef-5.3.0.jar!/chat2db-update-keys.properties
java -cp chat2db-studio-jcef.jar:chat2db-community-jcef.jar Probe
-> url=jar:...chat2db-studio-jcef-5.3.0.jar!/chat2db-update-keys.properties
```

Consequence for the commercial desktops: a Pro or Local installation would trust Community's key (or Community's unsubstituted placeholder) instead of its own, so manifests signed for Pro or Local would fail with "signing key is not trusted" and existing commercial users would lose automatic updates. This had to be fixed before Studio bumps `.chat2db-community-ref` past the rename.

## Fix

Community keeps a product specific resource name again and loads it explicitly, while the property contract stays unified: both products read `chat2db.update.key-id` / `chat2db.update.public-key`, which is what the shared `chat2db_update_java_options()` helper emits and what Studio's launcher options provide.

- `chat2db-community-jcef`: resource renamed to `chat2db-community-update-keys.properties` (placeholders unchanged), pom include/exclude updated, `GitHubReleaseDesktopUpdater` calls `TrustedUpdateKeys.load("/chat2db-community-update-keys.properties")`.
- `package-community-jcef.sh` and `script/package/README-updates.md` follow the name.

No Studio change is needed, and the Studio jar keeps its own resource.

## Verification

- `mvn -o -pl :chat2db-community-jcef process-resources -Dchat2db.update.key-id=PROBE-KEY -Dchat2db.update.public-key=PROBE-PUB` substitutes both values in the renamed resource; without `-D` the placeholders stay.
- Freshly packaged jar contains exactly one `update-keys` resource, the Community specific name, with the substituted values.
- Packaging guard probed with the real script: no key -> skip (exit 0); stale jar with an old resource -> `[error] bundled update signing key is not substituted` (exit 1); freshly built keyed jar -> `[check] bundled update signing key present` (exit 0). Probe artifacts restored.
- `mvn ... -Dtest='GitHubReleaseDesktopUpdaterTest,DesktopUpdaterRegistryTest,TrustedUpdateKeysTest,UpdateKeyIsolationTest'` -> 7 tests, 0 failures, BUILD SUCCESS.
- Repository search: the only remaining default resource reference is `TrustedUpdateKeys.BUNDLED_KEYS_RESOURCE`, used by products that do not pass an explicit resource (Studio). `git diff --check` clean.


openai0229:
Withdrawn: this change was not requested. The finding is recorded as a review item and the branch is kept for later.

